PR #115, opened 2019-05-13
adjust updates

dkim32:

**Description of Change:**
- replaced "paid attribution" with "non-organic" 
**Reason for Change:**

Convo here: 
https://brazetechnology.slack.com/archives/C221S39U5/p1557779594049900

i think we can adjust our language to simply say "organic" and "non-organic" given that there can still be attributable sources that may not be paid. a great example of that is attributing users to smart banners.

customers can attribute installs from the smart banner through an attribution partner. with adjust, they will need to follow the integration guidelines set here: https://docs.adjust.com/en/deeplinking/#tracking-safari-smart-banners-with-deep-links

Closes #**ISSUE_NUMBER_HERE**

---
---

## PR Checklist
- [ ] Ensure you have completed our CLA.
- [ ] Tag @EmilyNecciai as a reviewer when the your work is done and ready to be reviewed for merge. 
- [ ] Consult the [Docs Text Formatting Guide](https://github.com/Appboy/success/wiki/Docs-Text-Formatting-Guide) to be sure you're utilizing the proper markdown formatting.
- [ ] Consult the [Docs Writing Style Guide & Best Practices](https://github.com/Appboy/success/wiki/Writing-Style-Guide-&-Best-Practices) to be sure you're aligning with our voice and other style best practices.
- [ ] [Preview your deployed changes](https://homeslice.braze.com/docs) to confirm that none of your changes break production. Pay close attention to links and images.
- [ ] Tag others as Reviewers as necessary.
- [ ] If you have modified any links, be sure to add redirects to `config/nginx.conf.erb`.

---
---

### DIFF
--- a/_docs/_partners/technology_partners/advertising_technologies/attribution/adjust.md
+++ b/_docs/_partners/technology_partners/advertising_technologies/attribution/adjust.md
@@ -7,7 +7,7 @@ alias: /partners/adjust/
 
 [Adjust](https://www.adjust.com/) is a mobile attribution and analytics company that combines attribution for advertising sources with advanced analytics for a comprehensive picture of business intelligence.
 
-Adjust allows you to import non-organic attribution data to segment more intelligently within your lifecycle campaigns.
+Adjust allows you to import non-organic install attribution data to segment more intelligently within your lifecycle campaigns.
 
 ## Requirements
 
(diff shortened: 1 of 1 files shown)


EmilyNecciai:
@dkim32 Should it still say install attribution? I think we lose some SEO and clarification by removing "install". SO the phrase would be "non-organic install attribution."

dkim32:
> @dkim32 Should it still say install attribution? I think we lose some SEO and clarification by removing "install". SO the phrase would be "non-organic install attribution."

updated. also, adjust is also unique where we also allow customers to pass post-install events from adjust into braze. i didn't call it out in any of the intro as adjust is the only attribution vendor we allow this. might want to change in the future but i'm okay with leaving it out for now. thoughts?

EmilyNecciai:
@dkim32 I think not calling it out in the intro is fine for now. We can add a section on it later.

PR #118, opened 2019-05-14
AMPHTML

EmilyNecciai:

**Description of Change:**
- Added AMPHTML Hidden Page. (www.braze.com/docs/amphtml)

### DIFF
--- a/_docs/_hidden/private_betas/amphtml.md
+++ b/_docs/_hidden/private_betas/amphtml.md
@@ -1,0 +1,32 @@
+---
+nav_title: AMPHTML
+permalink: "/amphtml/"
+---
+
+# AMPHTML
+
+With AMPHTML, you can add interactive elements to your emails elevate your communications with your customers to a whole new level.
+
+## AMPHTML Email via API
+
+When you use any of [our Messaging Endpoints]({{ site.baseurl }}/api/endpoints/messaging/) to send an email, add `amp_body` as an object specification, as shown below.
+
+### Email Object Specification
+
+```json
+{
+  "app_id": (required, string) see App Identifier above,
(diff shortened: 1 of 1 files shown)


_docs/_hidden/private_betas/amphtml.md
```diff
@@ -0,0 +1,32 @@
...
+  "message_variation_id": (optional, string) used when providing a campaign_id to specify which message variation this message should be tracked under,
+  "extras": (optional, valid Key-Value Hash), extra hash - for SendGrid customers, this will be passed to SendGrid as Unique Arguments,
+  "headers": (optional, valid Key-Value Hash), hash of custom extensions headers. Currently, only supported for SendGrid customers,
+  "should_inline_css": (optional, boolean), whether to inline css on the body. If not provided, falls back to the default css inlining value for the App Group,
+  "attachments": (optional, array), array of json objects like [{"file_name","url"}] that define the files you need attached. Your file name's extension will be detected automatically from the URL, which should return the appropriate `Content-Type` as a response header,
+  "amp_body": (optional, updates the text-amp-html MIME type) the email body in AMPHTML. The MIME (Multipurpose Internet Mail Extensions) type to be referenced is "text/x-amp-html".
```
gsinghAppboy: only comment, and its really small and has no impact on the api call, would be to put amp_body before the 'body' value.  Also plaintext_body before amp_body.

So the order I would like to read this in should be plaintext_body, amp_body, and body.  It will make it consistent with how Google talks about sending the different MIME types.

EmilyNecciai: Great feedback! Going to apply all of these changes.

PR #146, opened 2019-05-29
Modernize ios custom endpoint docs

Bucimis:
1) The old integration examples are now broken due to us changing the default backend endpoint. 
2) We should not expose the backend endpoint as a public interface, which this old integration style does, so we should not present the old integration as an acceptable alternative.


Bucimis:
@zzhaobraze please merge

Bucimis:
@EmilyNecciai please merge

PR #178, opened 2019-06-19
CC Help Article - Guru

EmilyNecciai:

**Description of Change:**
Transfer Guru Card with Help Article Information. 
**Reason for Change:**
Alignment.

### DIFF
--- a/_docs/_help/help_articles/content_cards/empty_content_card_feed.md
+++ b/_docs/_help/help_articles/content_cards/empty_content_card_feed.md
@@ -7,6 +7,6 @@ page_order: 0
 
 You cannot send cards to users who do not exist. Therefore, by definition, new users will not have any cards in their feed on their first session.
 
-To ensure a new user receives cards, [create a Content Cards campaign]({{ site.baseurl }}/user_guide/message_building_by_channel/content_cards/overview/#content-cards) that is triggered on session start events (with no re-eligibility, so that users only receive the campaign once). They will receive this card on their next session. 
+If you want a new user to receive cards, [create a Content Cards campaign]({{ site.baseurl }}/user_guide/message_building_by_channel/content_cards/overview/#content-cards) that is triggered on session start events (with no re-eligibility, so that users only receive the campaign once). They will receive this card on their next session. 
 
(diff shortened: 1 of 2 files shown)


_docs/_help/help_articles/content_cards/empty_content_card_feed.md
```diff
@@ -0,0 +1,13 @@
+---
+nav_title: Empty Content Card Feed
+page_order: 0
+---
+
+# New Users Do Not Have Cards in Their Feed
+
+If brand new users, in their first session, do not have any cards in their feed you should...
```
toksfifo: Maybe something like:

> You cannot send cards to users who do not exist. Therefore, by definition, new users will not have any cards in their feed on their first session. One way to send cards to new users is to create a Content Cards campaign that is triggered on session start events (with no re-eligibility, so that users only receive the campaign once). 
>
> They will receive this card on their next session. Alternatively, to send the card in their first session, call `requestContentCardsRefresh` on the SDK during that session, a few seconds after calling `requestImmediateDataFlush`.

EmilyNecciai:
@toksfifo @hokstuff This is ready for a second look!

_docs/_help/help_articles/content_cards/empty_content_card_feed.md
```diff
@@ -0,0 +1,12 @@
+---
+nav_title: Empty Content Card Feed
+page_order: 0
+---
+
+# New Users Do Not Have Cards in Their Feed
+
+You cannot send cards to users who do not exist. Therefore, by definition, new users will not have any cards in their feed on their first session.
+
+To ensure a new user receives cards, [create a Content Cards campaign]({{ site.baseurl }}/user_guide/message_building_by_channel/content_cards/overview/#content-cards) that is triggered on session start events (with no re-eligibility, so that users only receive the campaign once). They will receive this card on their next session.
```
toksfifo: YMMV but IMO "To ensure" makes it sound like a recommendation rather than a suggestion. We're not necessarily telling customers to do this (rather, we're telling them that they can do this if they like).

EmilyNecciai: Valid! I made the change - thank you!

PR #233, opened 2019-07-09
Connected Content support in Content Cards

davidbielik:

**Description of Change:**

Removing Content Cards from list of unsupported channels

**Reason for Change:**

Content cards will now support Connected Content


EmilyNecciai:
Woo! :)

PR #237, opened 2019-07-11
SDK Changelog Image Fix

EmilyNecciai:
@zzhaobraze I can't get the images for Unity and Cordova to show - I think I'm doing something wrong here.


EmilyNecciai:
Thank you @zzhaobraze

PR #246, opened 2019-07-16
Add pinned icon instructions for android

Bucimis:


EmilyNecciai:
@Bucimis Is this something that can go out at anytime?

_docs/_developer_guide/platform_integration_guides/android/content_cards/customization.md
```diff
@@ -71,6 +71,12 @@ Here is a truncated example with a custom font family, `my_custom_font_family`,
   </style>
 ```
 
+### Setting A Custom Pinned Icon {#setting-a-custom-pinned-icon-for-android}
+
+To set a custom pinned icon, create a `png` image asset named `icon_pinned` and place it in your `drawable` folders. You will need to place an image in each `drawable` resolution you support, e.g. `drawable-hdpi`.
```
radixdev: I think it'd be more foolproof to override the style instead of the asset. See https://github.com/Appboy/appboy-android-sdk/blob/master/android-sdk-ui/src/main/res/values/styles.xml#L643-L649

Bucimis: good call - updated

_docs/_developer_guide/platform_integration_guides/android/content_cards/customization.md
```diff
@@ -71,6 +71,10 @@ Here is a truncated example with a custom font family, `my_custom_font_family`,
   </style>
 ```
 
+### Setting A Custom Pinned Icon {#setting-a-custom-pinned-icon-for-android}
+
+To set a custom pinned icon, override the `Appboy.ContentCards.PinnedIcon` style. Your custom image asset should be declared in the `android:src` element.
```
radixdev: I'd add a link OR a "see android-sdk-ui/styles" or something here.

Bucimis: I'm not sure this is necessary - the "Default Styling" section introduces the concept of overriding styles with instructions how to do it and the precending "Setting A Custom Font" section also already assumes the integrator knows how to override styles. It might be helpful to add a modified version of the existing style as an example though. Thoughts?

radixdev: Yeah I guess if we already link to the styles, then we're good here 👍

Bucimis:
@EmilyNecciai this is ready anytime!

PR #265, opened 2019-07-25
Service Worker registration details

davidbielik:
Added information for sites that already have a registered service worker

### DIFF
--- a/_docs/_developer_guide/platform_integration_guides/web/push_notifications/integration.md
+++ b/_docs/_developer_guide/platform_integration_guides/web/push_notifications/integration.md
@@ -36,9 +36,12 @@ This is a security requirement in the open standards specification that Braze We
 
 While industry best practice is to make your whole site secure, customers who cannot secure their site domain can work around the requirement by using a secure modal. Braze has prepared an [example of this approach][4].
 
-### Step 1: Configure your Site
+### Step 1: Configure your Site's Service Worker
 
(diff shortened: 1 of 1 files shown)


_docs/_developer_guide/platform_integration_guides/web/push_notifications/integration.md
```diff
@@ -36,9 +36,12 @@ This is a security requirement in the open standards specification that Braze We
 
 While industry best practice is to make your whole site secure, customers who cannot secure their site domain can work around the requirement by using a secure modal. Braze has prepared an [example of this approach][4].
 
-### Step 1: Configure your Site
+### Step 1: Configure your Site's Service Worker
+
+- If you don't already have a Service Worker, create a new file named ```service-worker.js``` with the content below, and place it in the root directory of your website.
+
+- Otherwise, if your site already registers a Service Worker, add the content below to the Service Worker file, and ensure that you specify its location using the [```serviceWorkerLocation``` initialization option](https://js.appboycdn.com/web-sdk/latest/doc/module-appboy.html#.initialize) when initializing the Web SDK.
```
froodian: I would say that the recommendation for people who already have a service worker is to use the pre-release `manageServiceWorkerExternally` option.  Providing `serviceWorkerLocation` doesn't really help all that much with an existing service worker, I don't think there's a great solution until we release that option.

EmilyNecciai:
@davidbielik Just wanted to see when this should be deployed - thanks!

PR #269, opened 2019-07-26
Sanitize algolia process to remove html tags

zzhaobraze:
Sanitize search script to remove the rendering of the query string as part of the results without removing html tags.

### DIFF
--- a/_includes/search_page.html
+++ b/_includes/search_page.html
@@ -82,7 +82,7 @@
         urlsearch += ' ' + currenturl.hash.replace('#', '');
       }
       if (urlsearch) {
-        $(searchbox).val(urlsearch.replace(/\+/g, ' ').replace(/\%20/g, ' ').replace(/\_/g, ' '));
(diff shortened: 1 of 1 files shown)


_includes/search_page.html
```diff
@@ -82,15 +82,14 @@ <h1 style="padding-top:40px;color: #212123;">Advanced Search </h1>
         urlsearch += ' ' + currenturl.hash.replace('#', '');
       }
       if (urlsearch) {
-        $(searchbox).val(urlsearch.replace(/\+/g, ' ').replace(/\%20/g, ' ').replace(/\_/g, ' '));
+        $(searchbox).val(urlsearch.replace(/\+/g, ' ').replace(/\%20/g, ' ').replace(/\_/g, ' ').replace(/</g,'').replace(/>/g,'').replace(/&lt;/g,'').replace(/&gt;/g,''));
```
jonhyman: Can we add lodash or underscore or something and use `_.escape` in the long run? It's possible this isn't the full set of things that you could XSS with.

zzhaobraze: Would need to include the lodash library just for that function, but according to the docs, it escapes "&", "<", ">", '"', and "'".
https://lodash.com/docs/#escape

I escape those also if you want.

jonhyman: oh, okay. cool yeah you should escape what it does.

PR #428, opened 2019-09-23
Content Cards in Canvas

davidbielik:


_docs/_user_guide/engagement_tools/canvas/create_a_canvas/content-cards_in_canvas.md
```diff
@@ -0,0 +1,71 @@
...
+
+### [Steps with Multiple Channels](#steps-with-multiple-message-channels)
+
+![content-card-multiple-channels.png][3]
+
+Canvas steps with a Content Card and another messaging channel have the following advancement options:
+
+| Option | Description |
+|---|---|---|
+| Message Sent | When message sent is selected, users must be successfully sent at least one of the message types in the step <br> <br>  _If the content card expires and the user hasn't been sent the email/webhook/push or viewed the content card, they will exit the Canvas and will not advance to subsequent steps._ |
```
MarionNammack: I think the details about expiry in this section are confusing, expiry is only relevant if the content card is sent, so I  would just use the explanation from the previous single-channel section

PR #507, opened 2019-10-03
BD-327 - updated bullet point formatting

davidbielik:
BD-327 - updated bullet point formatting


EmilyNecciai:
@davidbielik I'm not sure why my fix worked, but it did. 🤷‍♀

PR #534, opened 2019-10-15
Update openurl example

Bucimis:


_docs/_developer_guide/platform_integration_guides/ios/advanced_use_cases/linking.md
```diff
@@ -62,17 +62,34 @@ For more information, refer to [Apple's documentation][12] on the `LSApplication
...
 }
 ```
 
+{% endtab %}
+{% tab swift %}
+
+```swift
+func application(_ app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey : Any] = [:]) -> Bool {
+  let path = url.path
+  let query = url.query
```
hokstuff: To be consistent with the ObjC part, I think we should add [this comment](https://github.com/Appboy/braze-docs/pull/534/files#diff-3814ff61f7f80f182dbcf9437e9bdd74R74) in both

Bucimis: done 👍

Bucimis:
@EmilyNecciai please merge

_docs/_developer_guide/platform_integration_guides/ios/advanced_use_cases/linking.md
douglas-putnam: Swift is good. Separately, I'd replace {% tab swift %} with {% tab SWIFT %} and insert the same comment "// Here you should insert code to take some action based upon the path and query." as a new line under "let query = url.query"

PR #564, opened 2019-10-17
Update key value names, remove outdated opt-in statement

Bucimis:


Bucimis:
@EmilyNecciai please merge

PR #569, opened 2019-10-18
Remove references to iOS 8 and below

Bucimis:


EmilyNecciai:
@cla-bot check

EmilyNecciai:
@Bucimis Could you add some detail in response to this comment about why these references were removed? Thanks! :) (Just going to add it to our docs release summary)

Bucimis:
@EmilyNecciai the SDK dropped support for iOS 8 in January and XCode's current version is 11 (and integrators will almost certainly be using 9+), so references to older versions of iOS and XCode aren't useful.

PR #571, opened 2019-10-18
Make Carthage docs not dependent on outdated hardcoded version

Bucimis:
I imagine there could be a more elegant solution, but this is at least more future proof.


EmilyNecciai:
@Bucimis This is a great solution! We can chat about more scalable options at another time. :)

PR #575, opened 2019-10-18
Remove references to UIWebView

Bucimis:


_docs/_user_guide/message_building_by_channel/in-app_messages/customize.md
```diff
@@ -207,12 +207,6 @@ To use a video in iOS in-app messages, you may require developer assistance. You
...
 
-__UIWebView Settings__
-```
-webView.allowsInlineMediaPlayback = YES;
-webView.mediaPlaybackRequiresUserAction = NO;
-```
-
 __WKWebView Settings__
 ```
 WKWebViewConfiguration *config = [WKWebViewConfiguration new];
```
hokstuff: Double checking - but with how we have implemented WKWebView today, are these docs accurate with the current world?

Bucimis: Good point - I think it's actually irrelevant now (i.e. we have these settings on by default) - thoughts?

hokstuff: We have a slightly different default config [currently](https://github.com/Appboy/AppboyKit/blob/af5a5258331c7eeeebbcf2132d6d1cdb9d4f54ee/Appboy-iOS-SDK/AppboyUI/ABKInAppMessage/ViewControllers/ABKInAppMessageHTMLViewController.m#L36)

However, right now we don't have those settings set as customizable so it seems kind of misleading. And once we do allow for them to customize these settings, I think we'd want to add more beef to the docs corresponding to that new delegate

chshapiro: agreed - this whole section should be removed and once we add the delegate well write docs for that

Bucimis: I left a small note about default support. agree that we can add to it once the delegate is added

EmilyNecciai:
@Bucimis Looking for an update on this - is it good to merge? Thanks!

Bucimis:
@EmilyNecciai yep its ready!

EmilyNecciai:
@cla-bot check

zzhaobraze:
@cla-bot check

PR #648, opened 2019-11-07
Remove references to iOS versions before 3.0.0 + remove dry run reference

Bucimis:
Note that the docs no longer claim we use dry run or not based on the fact that both are theoretically possible.

### DIFF
--- a/_docs/_developer_guide/platform_integration_guides/ios/push_notifications/troubleshooting.md
+++ b/_docs/_developer_guide/platform_integration_guides/ios/push_notifications/troubleshooting.md
@@ -18,7 +18,7 @@ There are two types of provisioning profiles and certificates - development and 
 
 #### Step 2: Devices register for APNs and provide Braze with push tokens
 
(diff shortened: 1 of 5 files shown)


_docs/_developer_guide/platform_integration_guides/ios/push_notifications/troubleshooting.md
```diff
@@ -18,7 +18,7 @@ There are two types of provisioning profiles and certificates - development and
 
 #### Step 2: Devices register for APNs and provide Braze with push tokens
 
-When users open your app, they will be prompted to accept push notifications. If they accept this prompt, then APNs will generate a push token for that particular device. As of SDK version 2.21.0, the iOS SDK will immediately and asynchronously send up the push token for apps using the default [Automatic Flush Policy][40].  After we have a push token associated with a user, they will show as "Push Registered" in the dashboard on their user profile under the "Engagement" tab and will be eligible to receive push notifications from Braze campaigns.
+When users open your app, they will be prompted to accept push notifications. If they accept this prompt, then APNs will generate a push token for that particular device. The iOS SDK will immediately and asynchronously send up the push token for apps using the default [Automatic Flush Policy][40].  After we have a push token associated with a user, they will show as "Push Registered" in the dashboard on their user profile under the "Engagement" tab and will be eligible to receive push notifications from Braze campaigns.
```
hokstuff: Do we need to include the Flush Policy in this, since the token will be sent up in all cases? Not sure if it's necessary - thoughts?

EmilyNecciai: @hokstuff I would default to more information, as opposed to less, especially regarding mentioning Policies.

PR #690, opened 2019-11-20
[PC-1692] Add note for default image support in Xamarin

Bucimis:


Bucimis:
@EmilyNecciai please merge

hokstuff:
Ah okay, so the findings of that support ticket was that we don't support it on Xamarin 🧐

PR #691, opened 2019-11-21
[PC-1646] update action buttons docs

udayalla:
getActionWithIdentifier (to handle push notification action) which is currently deprecated, therefore we want to recommend users who use notification actions to implement the user notification framework.

The documentation should say to use didReceiveNotificationResponse:response IF users are using user notification framework for handling push. However, if not using user notification framework, getActionWithIdentifier is still required

### DIFF
--- a/_docs/_developer_guide/platform_integration_guides/ios/push_notifications/action_buttons.md
+++ b/_docs/_developer_guide/platform_integration_guides/ios/push_notifications/action_buttons.md
@@ -104,7 +104,7 @@ Appboy.sharedInstance()?.getActionWithIdentifier(identifier,
 {% endtabs %}
 
 
-> We strongly reccomend that people using `handleActionWithIdentifier` begin using UNNotification Framework.
+> We strongly recommend that people using `handleActionWithIdentifier` begin using UNNotification Framework.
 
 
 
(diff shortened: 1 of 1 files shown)


_docs/_developer_guide/platform_integration_guides/ios/push_notifications/action_buttons.md
```diff
@@ -80,6 +80,32 @@ Appboy.sharedInstance()?.userNotificationCenter(center,
...
+Appboy.sharedInstance()?.getActionWithIdentifier(identifier,
+                                                 forRemoteNotification: userInfo,,
+                                                 completionHandler: completionHandler)
+```
+
+{% endtab %}
+{% endtabs %}
+
+
+> We strongly recommend that people using `handleActionWithIdentifier` begin using UNNotification Framework.
```
Bucimis: Add that this is due to the deprecation of `handleActionWithIdentifier`

udayalla: done!

udayalla:
@EmilyNecciai 
Our docs reference `getActionWithIdentifier` (to handle push notification action) which is currently deprecated, therefore we want to recommend users who use notification actions to implement the user notification framework.

The documentation should say to use `didReceiveNotificationResponse:response` IF users are using user notification framework for handling push. However, if not using user notification framework, `getActionWithIdentifier` is still required.

PR #786, opened 2019-12-19
Clarify Avro File Write Thresholds

EmilyNecciai:

**Description of Change:**
> I'm clarifying the write threshold of Avro files for Currents partners. Let me know if I can clarify this further.


dulle-braze:
Hold off on this... I checked with the DI team, and I'm told the addition of `on the hour` isn't exactly accurate. Just curious... Where did this come from?

dulle-braze:
Yes, confirmed that `on the hour` isn't a timeframe during which we "cut" files and stream (this is already handled by the `every 5 minutes` timeframe. I'm not sure where this information came from, but my hypothesis is that it was confused with our partitioning mechanism, which creates new folders in the client's bucket for each hour's worth of events.

EmilyNecciai:
@dulle-braze "On the Hour" was always in docs in this file. I was just adding formatting to clarify it. I can get rid of that.

PR #920, opened 2020-02-12
Content Cards in Canvas GA

davidbielik:


davidbielik:
> "Information" misspelled
> 
> Will merge once fixed! Thanks!

Hi @KellieHawks we're actually removing that section :)

PR #965, opened 2020-02-26
Localization docs improvements including adding Czech and Ukrainian

Bucimis:
Need to release Czech support on all SDKs first.

Other changes:
1) Removes unnecessary info that could get out of date, like N languages and which exact strings are localized
2) Removes numbering from the languages lists to make future alphabetical additions easier

If someone has time:
1) Please consolidate into one list! SDKs will always update languages together


Bucimis:
@KellieHawks please merge!

PR #1024, opened 2020-03-24
BD-352, BD-446

KellieHawks:
BD-352 - External API encoding note
BD-446 - Updated data included in the exports
Grammar and spell check for entire developer guide


_docs/_developer_guide/platform_integration_guides/android/in-app_messaging/in-app_message_delivery.md
```diff
@@ -1,4 +1,4 @@
----
+hon---
```
zzhaobraze: this is causing error in the build

PR #1115, opened 2020-04-28
Incremental building of the documentation for debugging

radixdev:
**Why**
Local builds that used to take 60 seconds now take 4.

**Description of Change:**
Introduced incremental builds

The reason that https://github.com/Appboy/braze-docs/pull/605/files didn't work was because the `_config.yml` listed the documentation source directory as `.`, meaning that all files, including build temporary files, were treated as source by default and visible to the file watcher.

Thus, when the asset pipeline generated its css, js, etc. files, the file change watcher would pick them up as source changes, "regenerate" them and thus delete them. The end result was that those asset files would get deleted and break all the documentation pages. 

The _incremental_ fix in this PR was to move the asset pipeline folder out of the source tree. Since all files in this repo are technically source files, that meant moving the folder to whatever parent directory exists above where the documentation repo exists on the local machine.

The true proper fix would be to have all of the actual source files (like markdown pages, assets, includes, etc.), live in a `./src` folder and all of the files that are used to build the documentation be at root level.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -27,7 +27,13 @@ namespace :docs do
       puts `bundle exec jekyll build`
   end
   task :serve do
-    pipe 'bundle exec jekyll s --port 5006'
+    if ENV["SITE_URL"] == 'https://www.braze.com' && ENV["RACK_ENV"] == 'production'
+      pipe 'bundle exec jekyll s --port 5006'
+    else
+      # Force a clean build of the site and the pipeline assets
+      puts `rm .jekyll-metadata`
+      pipe 'bundle exec jekyll s --port 5006 --incremental'
+    end
   end
 end
 
--- a/Rakefile
+++ b/Rakefile
@@ -27,7 +27,7 @@ namespace :docs do
       puts `bundle exec jekyll build`
   end
   task :serve do
-    if ENV["USE_INCREMENTAL"]
+    if ENV["RACK_ENV"] == 'staging'
       # Force a clean build of the site and the pipeline assets
       puts `rm .jekyll-metadata`
       pipe 'bundle exec jekyll s --port 5006 --incremental --config _config.yml,_incremental_config.yml'
--- a/Rakefile
+++ b/Rakefile
@@ -27,12 +27,12 @@ namespace :docs do
       puts `bundle exec jekyll build`
   end
   task :serve do
-    if ENV["SITE_URL"] == 'https://www.braze.com' && ENV["RACK_ENV"] == 'production'
-      pipe 'bundle exec jekyll s --port 5006'
-    else
+    if ENV["USE_INCREMENTAL"]
       # Force a clean build of the site and the pipeline assets
       puts `rm .jekyll-metadata`
-      pipe 'bundle exec jekyll s --port 5006 --incremental'
+      pipe 'bundle exec jekyll s --port 5006 --incremental --config _config.yml,_incremental_config.yml'
+    else
+      pipe 'bundle exec jekyll s --port 5006'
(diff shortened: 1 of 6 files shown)


Rakefile
```diff
@@ -27,7 +27,13 @@ namespace :docs do
       puts `bundle exec jekyll build`
   end
   task :serve do
-    pipe 'bundle exec jekyll s --port 5006'
+    if ENV["SITE_URL"] == 'https://www.braze.com' && ENV["RACK_ENV"] == 'production'
+      pipe 'bundle exec jekyll s --port 5006'
+    else
+      # Force a clean build of the site and the pipeline assets
+      puts `rm .jekyll-metadata`
+      pipe 'bundle exec jekyll s --port 5006 --incremental'
```
radixdev, line 35: The enabling flag

_config.yml
```diff
@@ -191,6 +191,7 @@ asset_pipeline:
   output_path: assets
   display_path: docs/assets
   gzip: false
+  staging_path: '../.documentation_asset_pipeline'
```
radixdev, line 194: Moved the asset pipeline staging area to outside the source files

davidbielik:
image

zzhaobraze:
@radixdev one of the reason why incremental build isn't being used, is that if you edit the css or js files, jekyll doesn't update all the pages to point to the new compiled source codes, so you get broken js/css which will cause confusion for people editing docs and testing locally.
Though, restarting the server is an option.
@KellieHawks - what do you think of allowing this, but just making a note that if you edit the js/css(which 99.999% of people shouldn't be allow to change), you need to restart rakes?

radixdev:
@zzhaobraze Why remove the staging path? That will break the incremental building for the reasons I described earlier

zzhaobraze:
> @zzhaobraze Why remove the staging path? That will break the incremental building for the reasons I described earlier

it conflicts with the deploy on heroku and doesn't seem to be necessary.
``` Liquid Exception: Read-only file system @ dir_s_mkdir - /app/../.documentation_asset_pipeline in /_layouts/default.html```

**Comment by radixdev** [comment]
Ok. I'll change it to be behind an environment variable. But I can absolutely assure you it's necessary, for the reasons I described earlier

**Comment by KellieHawks** [comment]
> @radixdev one of the reason why incremental build isn't being used, is that if you edit the css or js files, jekyll doesn't update all the pages to point to the new compiled source codes, so you get broken js/css which will cause confusion for people editing docs and testing locally.
> Though, restarting the server is an option.
> @KellieHawks - what do you think of allowing this, but just making a note that if you edit the js/css(which 99.999% of people shouldn't be allow to change), you need to restart rakes?

@zzhaobraze 
I think it should be fine? If it does cause some issues we can always change it back. 
I'll update the github wiki to note this in a couple more places.

File: Rakefile [hunk 3]
```diff
@@ -27,7 +27,13 @@ namespace :docs do
       puts `bundle exec jekyll build`
   end
   task :serve do
-    pipe 'bundle exec jekyll s --port 5006'
+    if ENV["RACK_ENV"] == 'staging'
```
doboy, line 30: I think reversing the condition makes more sense here

i.e.

```
if ENV["RACK_ENV"] == 'development'
  do incremental
else
  standard serve
end
```

zzhaobraze: We wanted the default local to be incremental ie `rake` with an option to run it as non-incremental if needed if they're editing css/js ie `RACK_ENV=staging rake`. 
Hosting is done via `rake doc:build` so this only affects local usage.

PR #1145, opened 2020-05-07
Add connected content to content cards overview

davidbielik:


_docs/_user_guide/message_building_by_channel/content_cards.md
```diff
@@ -28,7 +28,7 @@ guide_featured_list:
 ---
 
 {% alert note %}
-Braze recommends that customers who use our News Feed tool move over to our Content Cards messaging channel - it is more flexible, customizable, and reliable. It is also easier to find and use in the Braze product. Contact your Braze account manager for more information.
+Braze recommends that customers who use our News Feed tool move over to our Content Cards messaging channel - it is more flexible, customizable, and reliable. It is also easier to find and use in the Braze product. See our [Migration Guide][/docs/user_guide/message_building_by_channel/content_cards/migrating_from_news_feed/] or contact your Braze account manager for more information.
```
KellieHawks, line 31: Please change link to:
(https://www.braze.com/docs/user_guide/message_building_by_channel/content_cards/migrating_from_news_feed/)

davidbielik: do we prefer to use relative links? it'll make it work when clicking in a local environment

KellieHawks: True, normally I would suggest including {{site.baseurl}} but inside alerts the format has to be this way or else the link doesn't load
Screen Shot 2020-05-07 at 11 00 24 AM

davidbielik: image
fixed! my mistake with the square brackets

PR #1277, opened 2020-06-26
Add 3.24.1 incompatibility note to Carthage integration instructions

Bucimis:


_docs/_developer_guide/platform_integration_guides/ios/initial_sdk_setup/carthage_integration.md
```diff
@@ -14,6 +14,10 @@ github "Appboy/Appboy-iOS-SDK" "<BRAZE_IOS_SDK_VERSION>"
 
 Make sure to replace `<BRAZE_IOS_SDK_VERSION>` with the latest version of the Braze iOS SDK in "x.y.z" format. Release versions are available [here](https://github.com/Appboy/appboy-ios-sdk/releases).
 
+{% alert note %}
```
Bucimis, line 17: We can revert this and rely on the changelog after releasing 3.24.2+, but right now, this is the latest version and may fail unexpectedly for folks who start integrating without specifying version - so we should give a heads up.

Bucimis:
We shouldn't merge this yet - still need to confirm behavior on 3.24.0

PR #1474, opened 2020-09-04
BD-656: Update Endpoints with new description field

KellieHawks:
BD-656: Update Endpoints with new description field


timothy-kim (approved):
👍

PR #1479, opened 2020-09-09
BD-668: Update Campaign_id field

KellieHawks:
BD-668: Update Campaign_id field in Sending Messages Immediately via API Only


timothy-kim (approved):
👍

PR #1503, opened 2020-09-16
BD-674, BD-665

KellieHawks:
BD-674 - Adding alias link to API blacklist doc
BD-665 - clarifying required SMS formatting


timothy-kim (approved):
👍

PR #1610, opened 2020-10-29
Update push_ttl_settings.md

davidbielik:
* removed outdated note on default TTL
* added note on what TTL does


chrishaines (approved):
Looks good. Do we want to also update the image because the copy is slightly different?

davidbielik:
> Looks good. Do we want to also update the image because the copy is slightly different?

Yep, i'll do that too, thanks

chrishaines:
> > Looks good. Do we want to also update the image because the copy is slightly different?
> 
> Yep, i'll do that too, thanks

Cool. The change is currently deploying to Sweeney so you might want to wait a few minutes for it to finish rolling out.

timothy-kim (approved):
👍

PR #1692, opened 2020-12-01
update changeUser to reference migration api

davidbielik:
* Our docs mention that external ID migration is not possible, so I've updated to reference our External ID Migration API Endpoint


KellieHawks (approved):
lgtm!

jonhyman (commented):
ty

PR #1771, opened 2021-01-20
clarify IDFA requirements

davidbielik:
clarified IDFA requirements so that it doesn't read like IDFA prompt is always required, rather it's required if you collect IDFA.


tbelote (approved):
I like the new text it is more clear.

Bucimis (approved):

KellieHawks (approved):
lgtm

PR #1679, opened 2020-11-22
SDK Auth [do not merge]

davidbielik:

PR #2017, opened 2021-04-08
SDK Authentication updates

davidbielik:
Todos:

- [ ]  Confirm SDK versions
- [ ]  Update iOS delegate examples


davidbielik (commented):
https://github.com/braze-inc/braze-docs/pull/2017/files#R275

> we should probably add more info here about the delegate - it's called ABKSdkAuthenticationDelegate. They need to create the delegate and implement its method handleSdkAuthenticationError: and then set the delegate using this method.
Once the code is live, we can only link to the delegate file in github

davidbielik:
https://github.com/braze-inc/braze-docs/pull/1679#discussion_r609218339

KellieHawks (approved):

PR #2110, opened 2021-05-11
[SDK-1513] Cleanup a bit of IAM/NF instructions

Bucimis:
Notably, remove assumption that integrators are on Cocoapods in IAM customization docs


chshapiro (approved):

_docs/_developer_guide/platform_integration_guides/ios/in-app_messaging/customization.md
```diff
@@ -17,102 +17,80 @@ __Note that integration of `SDWebImage` is required if you plan on using our Bra
...
-`ABKInAppMessage` objects may carry key value pairs as `extras`. These are specified on the dashboard when creating an in-app message campaign. Key value pairs can be used to send data down along with an in-app message for further handling by your app, allowing you to add custom behaviors on top of what Braze provides.
+`ABKInAppMessage` objects may carry key value pairs as `extras`. These are specified on the dashboard when creating a campaign. Key value pairs can be used to send data down along with an in-app message for further handling by your app.
 
 ## Setting Delegates
 
-In-app message display and delivery customizations can be accomplished in code by setting delegates. All of these customizations are entirely optional.
+In-app message display and delivery customizations can be accomplished in code by setting our optional delegates.
 
-### Core In-App Message Controller Delegate
+### In-App Message Delegate
```
hokstuff, line 26: nit: Would we want to call this `In-App Message UI Delegate`?

I feel that this naming may get confusing between `In-App Message Delegate` and `Core In-App Message Delegate`, since the former is just UI related

hokstuff (approved):
Nice cleanup, one suggestion for section naming

lowip (approved):

_docs/_developer_guide/platform_integration_guides/ios/in-app_messaging/customization.md
Bucimis: I disagree, for 95% of users this is the only IAM delegate they need and adding a qualifier makes things more confusing/verbose because it implies 1. other delegates with separate functionality exist somewhere else 2. that this delegate is only related to UI functions, which it's not

KellieHawks (approved):

PR #2183, opened 2021-05-27
Web SDK - Content-Security-Policy

davidbielik:
* Documenting CSP headers for Web SDK


_docs/_developer_guide/platform_integration_guides/web/csp_headers.md
```diff
@@ -0,0 +1,60 @@
...
+### connect-src
+
+- `connect-src https://sdk.iad-01.braze.com` - allows the SDK to communicate with Braze APIs.
+  - Change this URL to match your `baseUrl` initialization option's API Endpoint, as found [here](https://www.braze.com/docs/user_guide/administrative/access_braze/sdk_endpoints/).
+
+### script-src
+
+- `script-src https://js.appboycdn.com` - required when using the CDN-hosted integration.
+- `script-src 'unsafe-eval'` - required only when using the integration snippet which contains reference to `appboyQueue`
+  - To avoid using this directive, integrate using NPM or reference the `appboy.min.js` library directly via `<script src="path/to/appboy.min.js"></script>` without the `appboyQueue` logic.
```
wesleyorbin, line 41: This isn't true because the non-`npm` versions will still have `eval`. So you'd have to download through `npm` and then reference it like this or self-host it.

Also, we should include nonces here since they're required for HTML IAMs that have `<script>` blocks (which I assume most do).

_docs/_developer_guide/platform_integration_guides/web/csp_headers.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Font Awesome
+
+If you use Font Awesome icons within your dashboard campaigns, the following CSP directives are required:
+
+- `style-src https://use.fontawesome.com`
+- `style-src 'unsafe-inline'` or `style-src 'nonce-...'` are also required.
```
wesleyorbin, line 48: Maybe put the nonce option first?

wesleyorbin (approved):
🚀

_docs/_developer_guide/platform_integration_guides/web/csp_headers.md
davidbielik, line 48:
```suggestion
- `style-src 'nonce-...'` or `style-src 'unsafe-inline'` are also required.
```

_docs/_developer_guide/platform_integration_guides/web/csp_headers.md
```diff
@@ -0,0 +1,60 @@
...
+## Nonce Attributes
+
+If you use a `nonce` value in your `script-src` or `style-src` directives, pass that value to the `contentSecurityNonce` initialization option to propagate it to newly created scripts and styles generated by the SDK.
+
+```javascript
+import braze from "@braze/web-sdk";
+
+braze.initialize(apiKey, {
+  baseUrl: baseUrl,
+  contentSecurityNonce: "YWJjMTIz", // assumes a "nonce-YWJjMTIz" CSP value
```
froodian, line 26: might be better to replace this with `"YOUR-NONCE-HERE"` or something so everybody doesn't literally use `YWJjMTIz`

_docs/_developer_guide/platform_integration_guides/web/csp_headers.md
```diff
@@ -0,0 +1,60 @@
...
+
+### script-src
+
+- `script-src https://js.appboycdn.com` - required when using the CDN-hosted integration.
+- `script-src 'unsafe-eval'` - required only when using the integration snippet which contains reference to `appboyQueue`
+  - To avoid using this directive, integrate using NPM or reference the `appboy.min.js` library directly via `<script src="path/to/appboy.min.js"></script>` without the `appboyQueue` logic.
+
+## Font Awesome
+
+If you use Font Awesome icons within your dashboard campaigns, the following CSP directives are required:
```
froodian, line 45: `If you use Font Awesome icons within your dashboard campaigns` does this actually matter? I don't think it does? I think this may be necessary whether or not their campaigns actually use font awesome?  

Or maybe you're just saying that if they don't use them, they can just disable fontawesome as described below in "Alternatively"? If that's the case, I might reverse the order, since FA usage is on its way out anyway, so it's like, if you don't use fontawesome, use doNotLoadFontAwesome, alternatively, add the following CSP directives.

froodian (approved):

froodian:
Thanks for writing this up

_docs/_developer_guide/platform_integration_guides/web/csp_headers.md
davidbielik, line 26:
```suggestion
  contentSecurityNonce: "YOUR-NONCE-HERE", // assumes a "nonce-YOUR-NONCE-HERE" CSP value
```

_docs/_developer_guide/platform_integration_guides/web/csp_headers.md
davidbielik, line 41:
```suggestion
  - To avoid using this directive, integrate using NPM instead.
```

wesleyorbin (approved):

bre-fitzgerald (approved):

PR #2201, opened 2021-06-03
Glossary Capitalization

KellieHawks:
Ran through the Braze glossary and updated capitalization on many common Braze words like Content Cards, Canvas, in-app messaging, Currents, Push Stories, Liquid, News Feed, and more...


bre-fitzgerald (approved):
lgtm!

PR #2390, opened 2021-07-13
Update customization.md

mattlsteinberg:
Unwrap optionals in Swift code sample for setting `ABKInAppMessageUIDelegate`

https://www.braze.com/docs/developer_guide/platform_integration_guides/ios/in-app_messaging/customization/#in-app-message-delegate

Previous version of the code throws an error.

### Is this change associated with a Braze feature/product release?
- [ ] Yes (__Insert Feature Release Date Here__)
- [x] No


tbelote (approved):
LGTM, I agree that the example is obvious broken. I wonder if we want make a general pass at what should be nullable or what should be non null or whatever it is that guides the ? requirement.

Bucimis (approved):

chshapiro (approved):

KellieHawks (approved):

PR #2427, opened 2021-07-21
Cohort Import Section - Partner Docs

KellieHawks:
creating new cohort imports section in the partner docs to add misc partners that offer cohort imports, will redirect to the appropriate section if partner integration fits in an existing category.


bre-fitzgerald (approved):
lgtm!

PR #2611, opened 2021-09-01
Revert "BD-1060: Fix missing code snippet in push notification integration.md…"

bre-fitzgerald:
Reverts braze-inc/braze-docs#2609


bre-fitzgerald:
@KellieHawks looks like this content is not a code snippet, but a script that is not generating because the service is down. More info in Slack thread linked in the ticket.

KellieHawks (approved):

PR #2634, opened 2021-09-08
Add entry for anchor text to styling_test_page.md

bre-fitzgerald:

### DIFF
--- a/_docs/_home/styling_test_page.md
+++ b/_docs/_home/styling_test_page.md
@@ -50,7 +50,7 @@ You can link to headings with custom anchors by creating a standard link with a 
 
 {% raw %}
 ```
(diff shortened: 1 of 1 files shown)


KellieHawks:
Hey, @bre-fitzgerald this looks great! Can you include an example of how to reference the custom anchor? Something like: Here is my [link](#anchor-text). You might have to use {% raw %} or escape codes to get it to work.

KellieHawks (changes requested):
Hey @bre-fitzgerald can you change the second set of curly brackets to parenthesis. Thanks!

KellieHawks (approved):
Thanks for making those edits! Lgtm!

PR #2899, opened 2021-11-10
BD-1150: Update Catalogs with templating support

bre-fitzgerald:

Closes #**BD-1150**

- [x] Yes (__Insert Feature Release Date Here__)
- [ ] No

### DIFF
--- a/_docs/_hidden/private_betas/catalog.md
+++ b/_docs/_hidden/private_betas/catalog.md
@@ -134,7 +134,7 @@ Here's what this looks like when the Liquid is rendered:
 
 ![Example iOS push notification with catalog Liquid tags rendered][4]{: style="max-width:50%" }
 
-#### Templating catalog items with custom attributes
+#### Templating catalog items
 
 You can also use templating to dynamically pull catalog items based on custom attributes. For example, let's say a user has the custom attribute `wishlist`, which contains an array of game IDs.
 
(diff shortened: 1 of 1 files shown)


_docs/_hidden/private_betas/catalog.md
```diff
@@ -132,7 +132,45 @@ For example, to add the `image_link` from our Games catalog to our promotional m
 
 Here's what this looks like when the Liquid is rendered:
 
-![Example iOS push notification with catalog Liquid tags rendered][4]{: style="max-width:60%" }
+![Example iOS push notification with catalog Liquid tags rendered][4]{: style="max-width:50%" }
+
+#### Templating catalog items with custom attributes
```
johndinh, line 137: The templating is pretty generic and we can fill in the variable using custom event properties too.  

I'll tag @toksfifo for additional feedback around anything else we may want to call out here.

toksfifo: Yea you can template pretty much anything. I'd consider calling this just "Templating catalog items" or something of that ilk.

_docs/_hidden/private_betas/catalog.md
```diff
@@ -132,7 +132,45 @@ For example, to add the `image_link` from our Games catalog to our promotional m
...
+{% catalogs /catalogs/6171a881759044006998ed9a/items/{{ wishlist[0] }} %}
+
+Get {{item.title}} now, for just {{item.sale_price}}!
+```
+{% endraw %}
+
+Which will display as the following:
+> Get Tales now, for just 7.49 USD!
+
+You can use this format for adding images from your catalog, as mentioned in the [Using images](#using-images) section above.
```
toksfifo, line 172: YMMV, but I don't think we need to explicitly call out that this works for images — it works for any field that can be templated. For instance, you can use it to template an API request in the webhook composer.

toksfifo, line 172: I think it's important to call out the overall benefit somewhere in this templating section, perhaps right here. Something like:

> With templating, you can render a different catalog item for each user based on their individual custom attributes, event properties, or any other templatable field.

bre-fitzgerald:
@cla-bot check

toksfifo (approved):
🙌

PR #3024, opened 2021-12-15
Clarify email analytics

lydia-xie:
Updating Email Reporting and the Email Analytics Glossary to clarify "Machine opens" vs. "Other opens"


bre-fitzgerald (approved):
lgtm!

PR #3044, opened 2021-12-17
Remove mention of News Feed from tags.md

bre-fitzgerald:
Since we're moving away from News Feed and towards Content Cards, removing mention of News Feed as an example of content to use tags on.


lydia-xie (approved):
lgtm!

PR #3048, opened 2021-12-20
Android action button + Image aspect ratio note

Bucimis:


davidbielik (approved):
10/10 would approve again

bryanlogan (approved):
I'm wondering if the term "safe area" might be helpful for people searching.

chshapiro (approved):

Bucimis:
@bryanlogan "safe zone" is the term we use on the dashboard, but I'd like to avoid it here until we have more specific guidance and to help future-proof a bit.

bre-fitzgerald (approved):

PR #3049, opened 2021-12-21
Update key-value pairs to replace News Feed with Content Cards

bre-fitzgerald:
Since Content Cards is recommended over News Feed, replacing mentions of News Feed in the description and intro with Content Cards.


lydia-xie (approved):
lgtm!

PR #3119, opened 2022-01-20
removing specialties in partner search

KellieHawks:


KellieHawks:
@zzhaobraze when you get a chance can you fix the CSS please. I thought I could fix it but I still have a lot to learn about CSS.

PR #3129, opened 2022-01-24
API endpoint index 

bre-fitzgerald:
Adds an API endpoint index, replacing the current `config_only` endpoint page in the API docs. This page lists all available Braze endpoints grouped by category.


KellieHawks (approved):
Looks great!

PR #3138, opened 2022-01-25
update user alias links

davidbielik:
linking to alias docs https://www.braze.com/docs/user_guide/data_and_analytics/user_data_collection/user_profile_lifecycle/#user-aliases


bre-fitzgerald (approved):
lgtm!!

PR #3155, opened 2022-01-31
Update developer_nav.js

davidbielik:
but when you select Select Platform and then leave the page and come back, it shows up like this:

image (5)


davidbielik:
@zzhaobraze let me know if this fixes the issue you raised

zzhaobraze (approved):

KellieHawks (approved):

PR #3288, opened 2022-03-02
Fatal error code updates

KellieHawks:
Adding a warning and bolding text of the most common error code. This important code gets overlooked as it is listed last. This update will help solcons and io receive fewer support q's. co-signed by jamie


lydia-xie (approved):
lgtm!

PR #3412, opened 2022-04-05
Remove directional language: above pt.1

bre-fitzgerald:
Removing mentions of directional language (i.e., "above").

- When "above" refers to location within the UI, rework the sentence to avoid directional language
- When "above" refers to versions, use "later" instead
- When "above" refers to a section in the doc, use "preceding" or rework the sentence to avoid directional language


KellieHawks (approved):
Great edits!

PR #3504, opened 2022-04-28
BD-1367: Add year to sf_created_at note

KellieHawks:
@jayhahn @jeffreystutz 
We recently received a docs request asking for a clarification of the year on this note found in our snowflake docs:
"The value of `sf_created_at` is reliable only for events that were persisted after `Nov 15th, YEAR 9:31 pm UTC`."

Any help is much appreciated.


jayhahn (approved):
@KellieHawks added the year, thanks!

PR #3530, opened 2022-05-02
Fix incorrect page title for iOS

Bucimis:
In-App Message Templates -> Custom App Store review prompt


davidbielik (approved):

hokstuff (approved):
LGTM

And looking at it again, I think this page was intended to show examples of several IAM templates. But since there's only a single template, it makes sense to rename it like in this PR 👍

PR #4129, opened 2022-09-27
Updating user archival

KellieHawks:
Updating user archival page based on feedback from @sambellbraze in https://github.com/braze-inc/braze-docs/pull/4038


toksfifo:
Thanks for the ping @KellieHawks. I'm adding @braze-inc/core-objects to ack.

raghavnarain3 (approved):

PR #4386, opened 2022-11-09
BD-1733: Add info about integer data types being rounded at large numbers

bre-fitzgerald:


bre-fitzgerald:
@lydia-xie when reviewing, let me know if the structure that I introduced makes sense. I moved some things around because concepts weren't grouped together. We may want to consider breaking out the "Custom event properties" section into it's own page in the future because this page is quite long (possibly also the purchase event properties stuff as well). Thanks!

lydia-xie (approved):
lgtm!

PR #4458, opened 2022-11-29
Update put_update_existing_user_account.md

KellieHawks:
Duplication of name field in example request


josh-mccrowell-braze (approved):
Good catch!

PR #4493, opened 2022-12-05
2717 user management

josh-mccrowell-braze:
Ensured that all the information from the associated KB article was clearly articulated in this article


KellieHawks (approved):
Looks great! Minor fix to a spacing issue.

PR #4533, opened 2022-12-13
Fix redirects pointing at different locations for the same URL

bre-fitzgerald:

Closes #**ISSUE_NUMBER_HERE**

- [ ] No


KellieHawks (approved):

scott-braze:
I might be missing something, but the change here seems to be broken. If I visit <https://www.braze.com/docs/frequency_capping_tag>, then the result is a cryptic 404 page. If I visit <https://www.braze.com/docs/hidden/private_betas/frequency_capping_tag>, then I am redirected to <https://www.braze.com/docs/user_guide/engagement_tools/campaigns/testing_and_more/rate-limiting?redirected=true#frequency-capping>, which results in a Braze branded 404 message. If I delete the query string (`?redirected=true`), then it is the desired target: <https://www.braze.com/docs/user_guide/engagement_tools/campaigns/testing_and_more/rate-limiting#frequency-capping>

PR #4730, opened 2023-01-30
Defensive docs updates

josh-mccrowell-braze:
- Provided additional context around listed limitations
- Removed an SMS FAQ that doesn't seem like it's doing good work
- Removed the idiomatic phrase "out-of-the-box" from the docs


KellieHawks (approved):
Looks great!

PR #4732, opened 2023-01-30
Data Allowances Tweaks

KellieHawks:
- Added array saving data point method in tip box
- Moved non-billable data points before billable ones
- Removed "but there are nuances beyond what you see here which may affect the number of remaining points you expect to see."
- Looked in the docs for mentions of things that do not consume data points, added those to the list
- added block quote for most important data point sentences in first section


josh-mccrowell-braze (approved):
Looks great!

KellieHawks:
@josh-mccrowell-braze Fixed!

PR #5008, opened 2023-03-11
Feature flags FAQ

davidbielik:


_docs/_developer_guide/platform_wide/feature_flags.md
```diff
@@ -20,6 +20,9 @@ guide_featured_list:
 - name: Example Use Cases
   link: /docs/developer_guide/platform_wide/feature_flags/use_cases/
   fa_icon: fas fa-paint-brush
+- name: Frequently Asked Questions
+  link: /docs/developer_guide/platform_wide/feature_flags/faq/
+  fa_icon: fas fa-question
 ---
 
 To use feature flags, ensure your SDKs are up to date with at least these minimum versions:
```
josh-mccrowell-braze, line 28: I guess it's not a huge deal, but I don't exactly remember when this was added to the landing page. Prerequisites like this aren't _typically_ listed on the landing page. Any objections to me taking them out since they are in the articles themselves?

_docs/_developer_guide/platform_wide/feature_flags/use_cases.md
```diff
@@ -1,8 +1,9 @@
 ---
 nav_title: Example Use Cases
 article_title: Example Use Cases
-page_order: 3
+page_order: 30
```
josh-mccrowell-braze, line 4: I haven't yet grokked the rationale for not ordering the pages _specifically_ how you want them. I can sense you're following a best practice by using these larger numbers. Can you help me understand what you're doing?

_docs/_developer_guide/platform_wide/feature_flags/faq.md
```diff
@@ -0,0 +1,145 @@
...
+
+This is on our product roadmap. To help prioritize this, please raise this feedback with your Braze account team, or email our team: [feature-flags-feedback@braze.com](mailto:feature-flags-feedback@braze.com).
+
+## Technical Topics
+
+### How frequently are feature flags refreshed by the SDK? {#refresh-frequency}
+
+Feature flags are refreshed at session start, and when changing active users. Feature flags can also be manually refreshed using the SDK's [refresh method][refreshing].
+
+Keep in mind that refreshing feature flags may be throttled if performed too quickly (rate limiting subject to change), so it's best to only refresh before a user interacts with new features or periodically in the app if necessary.
```
josh-mccrowell-braze, line 96: I feel like I've asked this before, but do we have specifics about "may be throttled if performed too quickly"? Statements like this should be made with a light touch--either we should be very specific, or we should be very vague.

_docs/_developer_guide/platform_wide/feature_flags/faq.md
```diff
@@ -0,0 +1,145 @@
...
+
+### What happens if feature flags are refreshed mid-session? {#listen-for-updates}
+
+Feature flags may be refreshed mid-session. There are scenarios where you may want to update your app (if certain variables or configuration should change), while other scenarios you may not want to update your app (to avoid a shocking change in how your UI is rendered).
+
+To control this, you can [listen for updates][listen-for-updates] to feature flags and based on which feature flags have changed, make that determination to re-render your app or not.
+
+# Best Practices
+
+### Naming conventions
```
josh-mccrowell-braze, line 110: Consider if these best practices would be better placed in the "Create" article.

josh-mccrowell-braze (approved):
I've pushed some style changes to the branch and added a few questions. Once these small nits have been answered, I think this is good to go. Just ping me to merge.

_docs/_developer_guide/platform_wide/feature_flags.md
davidbielik: I thought it would be nice to be upfront with SDK versions on the landing page... I'm not opposed to removing it, but also would be in favor of adding it elsewhere :)

_docs/_developer_guide/platform_wide/feature_flags/use_cases.md
davidbielik: Yea, so just in case we want to add one between two pages, we don't need to re-order every subsequent page.

_docs/_developer_guide/platform_wide/feature_flags/faq.md
davidbielik: We don't, but it's meant as a vague warning that our backend can enforce or apply more strict rate limiting that may change over time.

_docs/_developer_guide/platform_wide/feature_flags/faq.md
davidbielik: Yea, that would be fine, maybe at the bottom of the create page?

_docs/_developer_guide/platform_wide/feature_flags.md
josh-mccrowell-braze: I'll talk to the team, but my intuition says that the principle of having a single source of truth that's referenced elsewhere would suggest stating this only once. I know that prerequisites probably won't change a lot, but I'm having to go back and replace all the iOS instances with Swift instances right now, so I know that they _can_ change. I'll remove for now.

_docs/_developer_guide/platform_wide/feature_flags/faq.md
josh-mccrowell-braze: I'm going to go ahead and merge this for now, and take a note to discuss with you further when you're off of PTO. We've been told by Solutions Consulting to treat these restrictions in a different way moving forward, and would like to finesse with you.

PR #5116, opened 2023-03-27
Created Live Activities FAQ

josh-mccrowell-braze:


bre-fitzgerald:
Hey @josh-mccrowell-braze, looks good! Can I ask, any reason why you went with adding a blank landing page instead of making the Live Activities article the parent page and nesting the FAQ underneath? It seems a bit odd to me to see **Live Activities** > **Live Activities**. I don't think we technically have a standard one way or the other, just curious. 

cc: @KellieHawks if you have any input on which way is preferred

josh-mccrowell-braze:
@bre-fitzgerald Nope no reason. I imagined there might be additional content eventually that might support the section. Happy to change it if that is a better practice!

bre-fitzgerald (approved):
Makes sense if the section might expand 👍 lgtm!

josh-mccrowell-braze:
@bre-fitzgerald 👍 If you'd prefer to have it in the different format until there's more articles, happy to reformat. Just let me know what you think is best. Thanks for looking at it!

PR #5165, opened 2023-03-30
[SDK-3595] Add BrazeConfig to disable location collection docs

Bucimis:
Updated:
https://www.braze.com/docs/developer_guide/platform_integration_guides/android/analytics/location_tracking#disabling-automatic-location-tracking

Not Updated:
https://www.braze.com/docs/developer_guide/platform_integration_guides/android/advanced_use_cases/locations_and_geofences/#step-3-enable-braze-location-collection
https://www.braze.com/docs/developer_guide/platform_integration_guides/android/initial_sdk_setup/android_sdk_integration/#step-5-enable-location-tracking

Reasoning: For enablement use-cases, using braze.xml will be simpler and less error prone. For disablement use cases, customers will often need a runtime option.

### DIFF
--- a/_docs/_developer_guide/platform_integration_guides/android/analytics/location_tracking.md
+++ b/_docs/_developer_guide/platform_integration_guides/android/analytics/location_tracking.md
@@ -32,13 +32,44 @@ With the release of Android M, Android switched from an install-time to a runtim
 
 ## Disabling automatic location tracking
 
-To disable automatic location tracking, set `com_braze_enable_location_collection` to false in `braze.xml`:
+#### Compile-time Option
+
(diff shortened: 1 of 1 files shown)


davidbielik (approved):

radixdev (approved):

Bucimis:
@bryanlogan interesting point, I'd say since the point of this PR is to add the runtime config (which `android-sdk-location` is not), we ship this narrow change now and separately evaluate holistically how we want to recommend people to turn on and off location functionality at compile time. It's possible we may want e.g. the presence of `android-sdk-location` code to replace `com_braze_enable_location_collection` completely, with `BrazeConfig` as a runtime override.

bryanlogan (approved):

josh-mccrowell-braze (approved):

PR #5364, opened 2023-05-04
noindex fix for config_only pages

KellieHawks:
Adding noindex so google search doesn't find the config_only pages


lydia-xie (approved):
lgtm!!!

PR #5421, opened 2023-05-12
Updated Customizing Feed

josh-mccrowell-braze:


_docs/_developer_guide/customization_guides/content_cards/customizing_feed.md
```diff
@@ -2,37 +2,251 @@
...
+{% tab Web %}
+
+Request a manual refresh of Braze Content Cards from the Web SDK at any time by calling [`requestContentCardsRefresh()`](https://js.appboycdn.com/web-sdk/latest/doc/modules/braze.html#requestcontentcardsrefresh). 
+
+You can also call [`getCachedContentCards`](https://js.appboycdn.com/web-sdk/latest/doc/modules/braze.html#getcachedcontentcards) to get all currently available cards from the last Content Cards refresh. 
+
+```javascript
+import * as braze from "@braze/web-sdk";
+
+function refresh(){
```
wesleyorbin, line 88: tiny formatting nit
```suggestion
function refresh() {
```

_docs/_developer_guide/customization_guides/content_cards/customizing_feed.md
```diff
@@ -54,49 +268,286 @@ Web content
...
+
+/**
+ * @param {String} feed_type - value of the "feed_type" KVP to filter
+ */
+function showCardsByFeedType(feed_type) {
+  braze.showContentCards(null, function(cards) {
+    return cards.filter(function(card) {
+      return card.extras["feed_type"] === feed_type;
+    });
+  })
```
wesleyorbin, line 531: Optional change but this version is a little more succinct
```suggestion
  braze.showContentCards(null, function(cards) {
    return cards.filter((card) => card.extras["feed_type"] === feed_type);
  });
```

wesleyorbin (approved):

spvismaya (approved):

_docs/_developer_guide/customization_guides/content_cards/customizing_feed.md
```diff
@@ -54,49 +268,286 @@ Web content
...
+When a user does not qualify for any Content Cards, an "empty feed" error message is displayed. 
 
-## Change "empty feed" language
-Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam convallis velit eget bibendum vulputate. Praesent sed mauris nec turpis faucibus bibendum vel et enim. Suspendisse potenti. Donec eleifend, arcu non molestie laoreet, mauris mauris sodales nunc, vitae euismod augue metus eu justo. Vivamus eleifend interdum ipsum, vitae hendrerit libero auctor sit amet. Fusce sodales ipsum sit amet risus venenatis ultricies. Etiam elementum risus vel lorem tincidunt varius. Sed euismod elit vel enim volutpat, quis dapibus mauris convallis. Sed nec quam a est tempor imperdiet id id ante. Sed in tortor vel libero placerat tincidunt.
+<!--Question: What does this look like? -->
+
+You can configure what is displayed in this empty feed message.
+
+![An empty feed error message that reads "This is a custom empty state message."][1]
```
radixdev, line 279: Looks like this on android: 
image

tbh we can probably just not have a picture here though.

radixdev (approved):

_docs/_developer_guide/customization_guides/content_cards/customizing_feed.md
```diff
@@ -54,49 +268,286 @@ Web content
...
+attributes.emptyStateMessage = "This is a custom empty state message"
+attributes.emptyStateMessageFont = .preferredFont(forTextStyle: .title1)
+attributes.emptyStateMessageColor = .secondaryLabel
 ```
+
+{% endsubtab %}
+{% subtab Objective-C %}
+
+<!-- Question: This content is pulled from the old iOS SDK. How can it be updated for the Swift SDK since customization through Attributes isn't supported for Objective-C? -->
+Change the language that appears automatically in empty Content Card feeds by [redefining the localizable content card strings](https://github.com/Appboy/appboy-ios-sdk/blob/3cca65b06f66085f5bc7c8e1ad267bf8bb1f0da7/AppboyUI/ABKContentCards/Resources/en.lproj/AppboyContentCardsLocalizable.strings) in your app's localizable strings file:
```
jerielng, line 321: It looks like we still have the same localizable assets in the Swift SDK [here](https://github.com/braze-inc/braze-swift-sdk/tree/main/Sources/BrazeUI/Resources/Localization/en.lproj). The `Attributes` empty state message pulls from these assets, so they should be able to do the same thing in an Objective-C context:
```
public var emptyStateMessage: String = localize(
  "braze.content-cards.no-card.text",
  for: .contentCard
)
```

josh-mccrowell-braze: Thanks! Could you please re-review this section just to make sure I got it right?

josh-mccrowell-braze:
@wesleyorbin : Thanks for giving this a look! Is there any option for **customizing displayed card order** or **changing “empty feed” language** for the Web SDK?

@jerielng : Thank you as well! Can you **customize displayed card order** in the Swift SDK?

wesleyorbin:
> @wesleyorbin : Thanks for giving this a look! Is there any option for **customizing displayed card order** or **changing “empty feed” language** for the Web SDK?

For customizing display order, you can use the `filterFunction` param of `showContentCards()`:

```javascript
braze.showContentCards(null, (cards) => {
  return sortBrazeCards(cards); // where sortBrazeCards is your sorting function that returns the sorted card array
});
```

There's no way to permanently change the empty feed language. You can replace it (example below), but you'd have to do it every time the feed is shown. However, the problem with that is the feed may take some time to refresh so the empty feed text won't be there right away. Overall, I wouldn't really recommend it.

```javascript
const emptyFeedText = document.querySelector(".ab-no-cards-message");
If (emptyFeedText) {
  emptyFeedText.innerHtml = "New empty feed language";
}
```

josh-mccrowell-braze:
@wesleyorbin : I took a basic swing at pulling those answers into the Web sections of the Customizing Feed article. Would you mind taking a look at those and correcting?

_docs/_developer_guide/customization_guides/content_cards/customizing_feed.md
josh-mccrowell-braze: Thanks! I agree about the picture. 

@radixdev Quick question: I'm trying to dig into what causes the empty feed message. I understand that the empty feed error message is a toast, and the linked picture is displayed when there’s no eligible feed. Updating  `EmptyContentCardsAdapter` only affects the view when there's no eligible feed, not the error displayed, correct? (I think they're the same message for Swift, so I'm trying to nail down the behavior here.)

_docs/_developer_guide/customization_guides/content_cards/customizing_feed.md
jerielng: Looks good! I'll take a look at the displayed order piece as well.

_docs/_developer_guide/customization_guides/content_cards/customizing_feed.md
```diff
@@ -49,54 +261,291 @@ Objective-C content
...
-"Appboy.content-cards.no-connection.title" = "Connection Error";
-"Appboy.content-cards.no-connection.message" = "Cannot establish network connection.\nPlease try again later.";
+
+{% endsubtab %}
+{% subtab Objective-C %}
+
+Change the language that appears automatically in empty Content Card feeds by redefining the localizable content card strings in your app's [`ContentCardsLocalizable.strings`]((https://github.com/braze-inc/braze-swift-sdk/tree/main/Sources/BrazeUI/Resources/Localization/en.lproj) file: 
+
+```obj-c
+public var emptyStateMessage: String = localize(
```
jerielng, line 325: There wouldn't be any Objective-C code here. In an Obj-C context, since they can't utilize the attributes struct, they would need to just update the value for the string in that `.strings` file for the corresponding locale.

wesleyorbin:
> @wesleyorbin : I took a basic swing at pulling those answers into the Web sections of the Customizing Feed article. Would you mind taking a look at those and correcting?

LGTM!

_docs/_developer_guide/customization_guides/content_cards/customizing_feed.md
```diff
@@ -2,37 +2,249 @@
...
+
+{% endsubtab %}
+{% endsubtabs %}
+
+The `ContentCardsFragment` source can be found on [GitHub](https://github.com/braze-inc/braze-android-sdk/blob/master/android-sdk-ui/src/main/java/com/braze/ui/contentcards/ContentCardsFragment.kt).
+
+{% endtab %}
+{% tab iOS %}
+
+<!--- Is this possible in the iOS SDK? If so, how? --->
```
jerielng, line 247: If they are using BrazeUI, in Swift, they can use the `.transform` method in the `Attributes` struct before presenting the BrazeUI Content Cards view controller.
```
var attributes = BrazeContentCardUI.ViewController.Attributes.defaults
attributes.transform = { cards in
    cards.sorted {
        if $0.pinned && !$1.pinned {
            return true
        } else if !$0.pinned && $1.pinned {
            return false
        } else {
            return $0.createdAt > $1.createdAt
        }
    }
}
let viewController = BrazeContentCardUI.ViewController(braze: AppDelegate.braze, attributes: attributes)
```

An Objective-C example might be a little more involved since we don't have access to the `attributes` struct. I'll circle back on that one.

jerielng: As far as Objective-C goes, there isn't really a way to do this without `attributes`. If a developer would like to sort the Content Cards, for example, they can do it using Swift and [bridge it to Objective-C](https://developer.apple.com/documentation/swift/importing-swift-into-objective-c). So we could say something like this in the Objective-C tab:
```
Customization via `BrazeContentCardUI.ViewController.Attributes` is not available in Objective-C. If you wish to customize the display of your Content Cards before presentation, you can create a separate file in Swift using the corresponding example and then call your Swift class from Objective-C.

For more details on bridging to Objective-C, refer to [Importing Swift into Objective-C](https://developer.apple.com/documentation/swift/importing-swift-into-objective-c).
```

_docs/_developer_guide/customization_guides/content_cards/customizing_feed.md
jerielng: Ah sorry I just realized the confusion here - the code snippet I attached above isn't an example of what they should do, but I was sharing it to show where that `"braze.content-cards.no-card.text"` string is being used in our SDK. I left a comment in the corresponding place, but basically, they would need to override the string for that key in their `Localizable.strings` file, rather than doing any Objective-C specific code.

_docs/_developer_guide/customization_guides/content_cards/customizing_feed.md
josh-mccrowell-braze: Thanks! This is very helpful.

We haven't, elsewhere, given workarounds like bridging from Swift to Objective-C, so I'm hesitant to do so here. If we wanted to provide this guidance elsewhere, we can do so, then link to it from our own documentation. Holding off on doing that until the need becomes apparent.

josh-mccrowell-braze:
@bre-fitzgerald : This is an article in the customization/content cards effort. It's being merged into a feature branch. Would you please give this a review for style and correctness any time before end of week? 

Let me know if you have any questions! Thank you very much!

_docs/_developer_guide/customization_guides/content_cards/customizing_feed.md
```diff
@@ -2,101 +2,557 @@
...
 
-### Step 1: Setting key-value pairs on cards
-Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam convallis velit eget bibendum vulputate. Praesent sed mauris nec turpis faucibus bibendum vel et enim. Suspendisse potenti. Donec eleifend, arcu non molestie laoreet, mauris mauris sodales nunc, vitae euismod augue metus eu justo. Vivamus eleifend interdum ipsum, vitae hendrerit libero auctor sit amet. Fusce sodales ipsum sit amet risus venenatis ultricies. Etiam elementum risus vel lorem tincidunt varius. Sed euismod elit vel enim volutpat, quis dapibus mauris convallis. Sed nec quam a est tempor imperdiet id id ante. Sed in tortor vel libero placerat tincidunt.
+### Step 1: Set key-value pairs on cards
+
+When creating a Content Card campaign, set [key-value pair data]({{site.baseurl}}/docs/developer_guide/customization_guides/content_cards/customizing_behavior/) on each card. Our filtering logic will use this key-value pair to categorize cards.
```
bre-fitzgerald, line 358:
```suggestion
When creating a Content Card campaign, set [key-value pair data]({{site.baseurl}}/developer_guide/customization_guides/content_cards/customizing_behavior/) on each card. Our filtering logic will use this key-value pair to categorize cards. 
```

_docs/_developer_guide/customization_guides/content_cards/customizing_feed.md
```diff
@@ -2,101 +2,557 @@
...
+  <item name="android:text">@string/com_braze_feed_empty</item>
+  <item name="android:textColor">@color/com_braze_content_card_empty_text_color</item>
+  <item name="android:textSize">18.0sp</item>
+  <item name="android:gravity">center</item>
+  <item name="android:layout_height">match_parent</item>
+  <item name="android:layout_width">match_parent</item>
+</style>
+```
+
+For more information on customizing Content Card style elements, see [Customizing style]({{site.baseurl}}/docs/developer_guide/customization_guides/content_cards/customizing_styles).
```
bre-fitzgerald, line 318:
```suggestion
For more information on customizing Content Card style elements, see [Customizing style]({{site.baseurl}}/developer_guide/customization_guides/content_cards/customizing_styles).
```

bre-fitzgerald (changes requested):
Content looks good to me! Just a few broken links

PR #5714, opened 2023-06-29
Replace Content Card creative details images

bre-fitzgerald:

Closes #**ISSUE_NUMBER_HERE**

- [ ] No


josh-mccrowell-braze:
This looks great! 

I can merge this right now and it would be nice lil spruce up of this article.

One other thing that has been on my list to tackle during the Content Card refresh: Technically, Braze provides five Content Card templates: Classic, Classic Image, Banner (to be renamed "Image"), Captioned Image, and Control. 

Control doesn't have a UI, and is mostly for devs. 

But a marketer - I believe - could interact with both Classic and Classic Image as mildly different experiences. Or at least, if they're the same, it would be good to say that they are mostly the same, but have differences on the back end. 

Is this step worth doing now?

bre-fitzgerald:
@josh-mccrowell-braze I think I'm a tad confused on this, is that to say that you want to differentiate between Classic Conetnt Cards having an image or not? If so, do you think it would be sufficient to say that the image is optional? I wouldn't call out the Control one here, as I think it would be confusing.

josh-mccrowell-braze:
@bre-fitzgerald : No, yeah, I agree about Control. 

Yes, I was asking if you thought it'd be worthwhile differentiating between Classic and Classic Image--or at least putting in a blurb that both exist as "templates."

bre-fitzgerald:
@josh-mccrowell-braze that seems somewhat incorrect from a UI perspective. The only options they have available in the dashboard are Classic, Captioned Image, and Banner. 


Differentiating between them as two templates on the backend seems like unneeded complexity from the marketer's side. We do also note that the image is optional.


Happy to discuss further if you have strong feelings about it but my suggestion would be to keep it as it 🤔

josh-mccrowell-braze:
Roger. I'll merge this version in, then!

josh-mccrowell-braze (approved):

PR #5785, opened 2023-07-13
Consolidate: Creating a Push Message

rachel-feinberg:


bre-fitzgerald (changes requested):
Looks good!! One thing, can we remove the "Notification Headers" tile from the parent page?

bre-fitzgerald (approved):

PR #5837, opened 2023-07-21
Consolidate push best practices

bre-fitzgerald:
Turns the landing page for "Push best practices" into a consolidated page of some of the shorter, related push best practices articles. Also folds in best practices from the "Push enablement and subscriptions" article.


rachel-feinberg (approved):
Lgtm!

PR #5874, opened 2023-07-27
Added a note about allowUserSuppliedJavascript to Android and Swift docs

josh-mccrowell-braze:
Received feedback that the `allowUserSuppliedJavascript` config option wasn't called out in the SDK docs (only the user guide). Added it in for Android and Swift. 

Is this config option rendered correctly for these platforms? 

Are there other platforms that we should be explicit about this for?


lowip:
@josh-mccrowell-braze, this is a Web SDK only setting and is not needed for both iOS and Android. 
I think we can close this PR.

josh-mccrowell-braze:
@lowip Thanks for clarifying! I'm having trouble understanding this feedback then. The Web SDK _does_ have this setting called out in the docs. I'll chalk it up to user error, I guess.

PR #5923, opened 2023-08-03
BD-2339 Update conditional logic

rachel-feinberg:

Closes [#BD-2339
](https://jira.braze.com/browse/BD-2339)
- [x] No


_docs/_user_guide/personalization_and_dynamic_content/liquid/conditional_logic.md
```diff
@@ -124,6 +127,20 @@ Note that a null attribute value isn't strictly associated with a value type (e.
 
 {% endraw %}
 
+### Blank attribute values
+
+A blank value occurs when the attribute on a user profile either isn't set or is set with an empty string. Blank values should be checked before other variables.
```
willng-braze, line 132: Feel free to word as you see fit but blank here is probably ruby language evaluation of blank.  That means nil/null PLUS all the empty values.  https://apidock.com/rails/Object/blank%3F

Definition here is pretty accurate, I think you're missing false

willng-braze (approved):
Wow TIL!  The liquid guide never mentions blank anywhere so I thought it was an alias for nil/null but it is different.

_docs/_user_guide/personalization_and_dynamic_content/liquid/conditional_logic.md
```diff
@@ -124,6 +127,20 @@ Note that a null attribute value isn't strictly associated with a value type (e.
 
 {% endraw %}
 
+### Blank attribute values
+
+A blank value occurs when the attribute on a user profile isn't set, is set with an whitepace string (i.e., ` ` ), or is set as `false`. Blank values should be checked before other variables to avoid a Liquid processing error.
```
bre-fitzgerald, line 132:
```suggestion
A blank value occurs when the attribute on a user profile isn't set, is set with a whitespace string (i.e., ` ` ), or is set as `false`. Blank values should be checked before other variables to avoid a Liquid processing error.
```

_docs/_user_guide/personalization_and_dynamic_content/liquid/conditional_logic.md
```diff
@@ -92,9 +92,13 @@ The `{% endif %}`  tag signals that you've finished your conditional logic. You
...
 
-## Accounting for null attribute values
+## Accounting for null, nil, and blank attribute values
 
-Conditional logic is a useful way to account for null attribute values. A null value occurs when the value of a custom attribute has not been set. For example, a user who has not yet set their first name will not have a first name in Braze's database.
+Conditional logic is a useful way to account for attribute values that aren't set in user profiles.
+
+### Null and nil attribute values
+
+A null or nil value occurs when the value of a custom attribute has not been set. For example, a user who has not yet set their first name will not have a first name in Braze's database.
```
bre-fitzgerald, line 101: Style guide edit: not using "Braze" as a possessive
```suggestion
A null or nil value occurs when the value of a custom attribute has not been set. For example, a user who has not yet set their first name will not have a first name logged in Braze.
```

bre-fitzgerald (changes requested):
looks great! just 2 minor comments for ya

bre-fitzgerald (approved):

PR #5926, opened 2023-08-04
404

davidbielik:
moved the custom event to an `else` only if no redirect exists to avoid false positives / race condition


zzhaobraze (approved):

zzhaobraze:
@davidbielik let the team know if you want this merged

josh-mccrowell-braze (approved):

PR #5929, opened 2023-08-07
BD-2320: Updates for email validation

lydia-xie:

Closes #**BD-2320**

- [ ] No

---

### DIFF
--- a/_docs/_user_guide/message_building_by_channel/email/email_setup/email_validation.md
+++ b/_docs/_user_guide/message_building_by_channel/email/email_setup/email_validation.md
@@ -13,31 +13,24 @@ channel: email
 
 > This reference article covers local and host part validation rules for email addresses.
 
-Validation is used for dashboard email addresses, end-user email addresses (your customers), and from and reply-to addresses done of an email message. Email validation is performed when a user's email address has been updated or is being imported into Braze via API, CSV Upload, SDK, or modified in the dashboard. Note that your email addresses cannot include white spaces, and if send using the API, whitespaces will result in a `400` error.
+Validation is used for dashboard email addresses, end-user email addresses (your customers), and from and reply-to addresses done of an email message. Email validation is performed when a user's email address has been updated or is being imported into Braze via API, CSV upload, or SDK, or modified in the dashboard. Note that your email addresses cannot include whitespaces. If you're using the API, whitespaces will result in a `400` error.
 
-Email addresses targeted via the Braze servers must be validated per [RFC 2822](https://datatracker.ietf.org/doc/html/rfc2822) standards, Braze does not accept certain characters and recognizes them as invalid. For example, email addresses cannot start or end with any special characters, including apostrophes. If an email is bounced, Braze marks the email as invalid and the subscription status is not changed.  
+Email addresses targeted via the Braze servers must be validated per [RFC 2822](https://datatracker.ietf.org/doc/html/rfc2822) standards. Braze does not accept certain characters and recognizes them as invalid. For example, email addresses cannot start or end with any special characters, including apostrophes. If an email is bounced, Braze marks the email as invalid and the subscription status is not changed.  
 
 {% details Unaccepted characters outside of RFC Standards %}
-- *
 - /
-- ?
-- !
 - $
-- #
-- %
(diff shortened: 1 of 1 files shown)


_docs/_user_guide/message_building_by_channel/email/email_setup/email_validation.md
```diff
@@ -13,31 +13,24 @@ channel: email
...
 
-Email addresses targeted via the Braze servers must be validated per [RFC 2822](https://datatracker.ietf.org/doc/html/rfc2822) standards, Braze does not accept certain characters and recognizes them as invalid. For example, email addresses cannot start or end with any special characters, including apostrophes. If an email is bounced, Braze marks the email as invalid and the subscription status is not changed.  
+Email addresses targeted via the Braze servers must be validated per [RFC 2822](https://datatracker.ietf.org/doc/html/rfc2822) standards. Braze does not accept certain characters and recognizes them as invalid. For example, email addresses cannot start or end with any special characters, including apostrophes. If an email is bounced, Braze marks the email as invalid and the subscription status is not changed.
```
jteich, line 18: I think that we should remove the reference to RFC-2822. We aren't quite true to that RFC and it any case we target an updated standard (RFC-5322 and RFC-6531). However, it's best not to mention any RFC since we don't entirely follow the RFC as documented here: https://github.com/Appboy/platform/blob/bb70346fb2cb64d8a0efbdaa62034112d4f49488/shared_code/lib/shared/appboy/communication/email_address_validator.rb#L23
Alternatively, you can mention those RFCs together with the exclusions, but I think it's best not to mention any RFC.

jteich: I just checked and realized that for our purposes, RFC-2822. and RFC-5322 are the same, but the overall idea still holds.

_docs/_user_guide/message_building_by_channel/email/email_setup/email_validation.md
```diff
@@ -13,31 +13,24 @@ channel: email
...
 
 {% details Unaccepted characters outside of RFC Standards %}
-- *
 - /
-- ?
-- !
 - $
-- #
-- %
 - &#94;
```
jteich, line 23: `&#94;` is the carat (^) which is valid and should be removed from this list. 
Do we actually need this list? It isn't comprehensive and and only includes ascii chars. For example, double-quotes are not in this list and are not valid.

_docs/_user_guide/message_building_by_channel/email/email_setup/email_validation.md
jteich: The sentence "For example, email addresses cannot start or end with any special characters, including apostrophes." isn't correct. The local part of an email address may start and end with any of the allowed characters except for a period.

_docs/_user_guide/message_building_by_channel/email/email_setup/email_validation.md
```diff
@@ -13,33 +13,45 @@ channel: email
...
+	- ~
+	- !
+	- ?
+	- . (only between letters other characters)
 {% enddetails %}
 
-This validation is not to be confused with a validation service like Briteverify. This is a check to verify that the syntax of an email address is correct. One of the main drivers to use this new validation process is to support international characters (i.e., UTF-8) in the local part of the email address.
+{% details Unaccepted characters %}
+- Whitespaces (ASCII and Unicode)
+- Only allowing Unicode
```
jteich, line 45: Should the description of allowed unicode characters be in the "Unaccepted characters" section?

_docs/_user_guide/message_building_by_channel/email/email_setup/email_validation.md
```diff
@@ -13,33 +13,45 @@ channel: email
...
+	- /
+	- %
+	- *
+	- =
+	- `
+	- |
+	- ~
+	- !
+	- ?
+	- . (only between letters other characters)
```
jteich, line 40:
``` suggestion
	- . (only between letters or other characters)
```

bre-fitzgerald (approved):

PR #6053, opened 2023-08-29
Update security_qualifications.md

bre-fitzgerald:
Updating comment to be more explicit, since this is the one disclosures file we edit semi-routinely.


rachel-feinberg (approved):
Lgtm!

PR #6074, opened 2023-08-31
Message Activity Log

davidbielik:
renamed Developer Console Log to match dashboard "Message Activity Log".

Might exist elsewhere too


bre-fitzgerald (approved):
Thanks for catching these!

PR #6097, opened 2023-09-06
Task-based edit: Push Action Buttons

rachel-feinberg:

Closes #**ISSUE_NUMBER_HERE**

- [x] No

### DIFF
--- a/_docs/_user_guide/message_building_by_channel/push/advanced_push_options/push_action_buttons.md
+++ b/_docs/_user_guide/message_building_by_channel/push/advanced_push_options/push_action_buttons.md
@@ -13,45 +13,47 @@ channel:
 
 ![An iOS push notification with two push action buttons: Accept and Decline.][1]{: style="float:right;max-width:40%;margin-left:15px;"}
 
-> Push action buttons enable you to set content and actions for buttons when utilizing Braze's iOS and Android push notifications. With action buttons, your users can interact directly with your app from a notification without needing to click into an app experience to take action.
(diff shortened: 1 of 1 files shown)


_docs/_user_guide/message_building_by_channel/push/advanced_push_options/push_action_buttons.md
```diff
@@ -13,45 +13,47 @@ channel:
...
-Then select your **Notification Category**. You can select from the following available button combinations:
-
-- Accept / Decline
-- Yes / No
-- Confirm / Cancel
-- More
-- Pre-registered Custom Category
+1. Create a [push campaign]({{site.baseurl}}/user_guide/message_building_by_channel/push/creating_a_push_message/) and enable action buttons in the **Compose** tab.
+2. Select your **iOS Notification Category** from the following available button combinations:
+ - Accept / Decline
```
rachel-feinberg, line 28: @bre-fitzgerald Should this list of text be bolded? It's part of the UI (dropdown), so I assume yes?

bre-fitzgerald: These I think we can leave as is. A, because it's just easier to read a list where everything isn't bolded and B, I believe we don't bold dropdown options unless we're telling them to select a specific item from the list.

So if we were saying "Select **Yes / No** for your **iOS Notification Category**" then we would bold it.

bre-fitzgerald (approved):

PR #6132, opened 2023-09-13
Liquid use case edits pt.2

rachel-feinberg:

Closes #**ISSUE_NUMBER_HERE**

- [ ] No

### DIFF
--- a/_docs/_user_guide/personalization_and_dynamic_content/liquid/liquid_use_cases.md
+++ b/_docs/_user_guide/personalization_and_dynamic_content/liquid/liquid_use_cases.md
@@ -649,7 +649,8 @@ Did you forget something in your shopping cart?
 ```
 {% endraw %}
 
-{% alert important %} You must have an event property of the custom event count or use a webhook to your Braze endpoint. This is to increment a custom attribute (`example_event_count`) every time the user performs the event. This example uses a cadence of three (1, 4, 7, 10, etc.).{% endalert %}
+{% alert important %} You must have an event property of the custom event count or use a webhook to your Braze endpoint. This is to increment a custom attribute (`example_event_count`) every time the user performs the event. This example uses a cadence of three (1, 4, 7, 10, etc.). To start the cadence from zero (0, 3, 6, 9, etc.), remove `minus: 1`.
+{% endalert %}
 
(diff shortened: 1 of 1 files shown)


_docs/_user_guide/personalization_and_dynamic_content/liquid/liquid_use_cases.md
```diff
@@ -1078,13 +1079,6 @@ This use case checks if a custom attribute array contains a specific string, and
 ```liquid
 {% if custom_attribute.${PartnershipProgramsNotLinked} contains 'Hertz' %}
 Link your Hertz account to use Hertz Fast Lane.
-
-{% elsif custom_attribute.${airportCompleted} == false %}
-Clear helps you breeze through airport security. Complete your one-time in-person setup next time you are at the airport. It only takes about 5 minutes.
-
-{% else %}
-Your account is all setup
-{% endif %}
```
bre-fitzgerald, line 1087: For my understanding, what is incorrect/inaccurate about this section of the Liquid? I believe deleting the `endif` statement will break this.

rachel-feinberg: The use case tester suggested to remove that portion as it's not specific to the example (finding a string in an array)

It’s fine but I would include this part in the example (the other parts don’t really apply to the example at hand). Can just close the condition statement after that if statement: 
{% if custom_attribute.${PartnershipProgramsNotLinked} contains 'Hertz' %}
Link your Hertz account to use Hertz Fast Lane.

bre-fitzgerald: Ah okay, I believe you'll want to add the `{% endif %}` statement back in then, as that would be closing the condition statement.

bre-fitzgerald: It's like in the alerts we use in docs (which also use Liquid), if you add `{% alert note %}` but don't include `{% endalert %}`, the whole thing goes 💥

rachel-feinberg: Ahhh that's for catching that

bre-fitzgerald (approved):

PR #6165, opened 2023-09-18
Liquid use case pt.4

rachel-feinberg:

Closes #NONE

- [x] No

### DIFF
--- a/_docs/_user_guide/personalization_and_dynamic_content/liquid/liquid_use_cases.md
+++ b/_docs/_user_guide/personalization_and_dynamic_content/liquid/liquid_use_cases.md
@@ -691,7 +691,7 @@ This use case calculates the number of times a custom event has been logged betw
 
 {% for custom_event in response.users[0].custom_events %}
 {% assign ce_name = custom_event.name %}
-{% comment %} The following Custom Event name will need to be amended for the target Custom Event. {% endcomment %}
+{% comment %} The following custom event name will need to be amended for the target custom event. {% endcomment %}
 
 {% if ce_name == "Project Exported" %}
 {% comment %}{{custom_event.name}}: {{custom_event.count}}{% endcomment %}
--- a/_docs/_user_guide/personalization_and_dynamic_content/liquid/liquid_use_cases.md
+++ b/_docs/_user_guide/personalization_and_dynamic_content/liquid/liquid_use_cases.md
@@ -681,7 +681,7 @@ This use case calculates the number of times a custom event has been logged betw
 
 {% endcapture %}
 
-{% connected_content <<Braze endpoint>>/users/export/ids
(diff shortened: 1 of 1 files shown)


_docs/_user_guide/personalization_and_dynamic_content/liquid/liquid_use_cases.md
```diff
@@ -674,6 +672,23 @@ This use case calculates the number of times a custom event has been logged betw
...
+
+{% capture body %}
+{
+ "braze_id": "{{${braze_id}}}",
+ "fields_to_export": ["custom_events"]
+}
+
+{% endcapture %}
+
+{% connected_content <<Braze endpoint>>/users/export/ids
```
bre-fitzgerald, line 684: Per style guide, placeholders are all caps

```suggestion
{% connected_content YOUR_BRAZE_ENDPOINT/users/export/ids
```

_docs/_user_guide/personalization_and_dynamic_content/liquid/liquid_use_cases.md
```diff
@@ -674,6 +672,23 @@ This use case calculates the number of times a custom event has been logged betw
...
+ :method post
+  :headers { "Authorization": "Bearer <<api_key>>" }
```
bre-fitzgerald, line 686:
```suggestion
  :headers { "Authorization": "Bearer YOUR_API_KEY" }
```

_docs/_user_guide/personalization_and_dynamic_content/liquid/liquid_use_cases.md
```diff
@@ -674,6 +672,23 @@ This use case calculates the number of times a custom event has been logged betw
...
+  :body {{body}}
+ :content_type application/json
+ :save response
+  :retry %}
+
 {% for custom_event in response.users[0].custom_events %}
 {% assign ce_name = custom_event.name %}
 {% comment %} The following Custom Event name will need to be amended for the target Custom Event. {% endcomment %}
```
bre-fitzgerald, line 694:
```suggestion
{% comment %} The following custom event name will need to be amended for the target custom event. {% endcomment %}
```

bre-fitzgerald (changes requested):
Looks good! Just some notes on standards. 

This is moderately new territory standards-wise, as we typically don't include placeholder text within Liquid examples. However, I did a quick search and for Connected Content it looks like we follow a similar approach to API docs, wherein our placeholders are in the format of YOUR_THING_NAME without any extra styling like { } or < > encapsulating it. 

This approach makes sense to me, as I could see the extra characters potentially being construed as part of the CC syntax. Regardless I think we should chat rq with the team about this standard and add it to the style guide.

bre-fitzgerald (approved):

PR #6317, opened 2023-10-10
Update font size of API

bre-fitzgerald:

#### Description of Change:
Updates font size of pages using API format (glossaries, API references) to use the same font size as the rest of docs (16px) rather than a smaller size (14px). This does not change the size of API methods or URLs.


josh-mccrowell-braze:
Nice! Thanks for doing this.

PR #6442, opened 2023-10-31
Update how DNS records are provided

lydia-xie:

Closes #**ISSUE_NUMBER_HERE**

- [ ] No

---


tiffanyjlin (commented):
I would suggest deleting this text and the screenshot below it because it looks like it's from Sendgrid, which our customers don't have access to:
To find the address, go to your SendGrid DNS record and copy the UID, Whitelabel Subdomain, and Domain from the Host Value column.

bre-fitzgerald (approved):

PR #6522, opened 2023-11-15
BD-2602: Changing an outdated link and sneaking in a lil test while I'm toodling

josh-mccrowell-braze:


rachel-feinberg (approved):
Lgtm!

PR #6717, opened 2024-01-10
Added small clarification on how the listener behaves in React

josh-mccrowell-braze:


josh-mccrowell-braze:
Can someone on the @braze-inc/sdks-wrappers team give this a quick review to make sure I'm correct about this, please?

_docs/_developer_guide/platform_integration_guides/react_native/content_cards.md
```diff
@@ -20,15 +20,15 @@ The Braze SDKs include a default card feed to get you started with Content Cards
 To build your own UI, you can get a list of available cards, and listen for updates to cards:
 
 ```javascript
-// set initial cards
+// Set initial cards
 const [cards, setCards] = useState([]);
 
-// listen for updates as a result of card refreshes
+// Listen for updates as a result of card refreshes: listener is called whenever the card is updated, such as on a new session, on a manual refresh with `requestContentCardsRefresh()`, or when a user calls up the Content Card feed and it's after the timeout period
```
hokstuff, line 26:
```suggestion
// Listen for updates as a result of card refreshes. The listener is called whenever the card is updated, such as on a new session, on a manual refresh with `requestContentCardsRefresh()`, or after the timeout period
```

I made some edits bc I wasn't sure what the phrase was referencing: `when a user calls up the Content Card feed` → Do we mean "pulls down on the feed"?

hokstuff (approved):
One suggestion, LGTM!

josh-mccrowell-braze:
Thanks @hokstuff !

@internetisaiah This is ready for a writer review and merge, please!

internetisaiah:
**Note:** I trimmed the line down, then split across two lines. (I don't think we have specific standards for this yet, but I was taught the soft-limit for line length should be 80 characters and the hard-limit for line length should be 125, per my C/C++ classes).

internetisaiah (approved):
LGTM

PR #6723, opened 2024-01-11
BD-2733 Compliance Center (to deploy on March 1st)

josh-mccrowell-braze:
**For deployment on March 1st**


josh-mccrowell-braze:
The Legal team wants to be able to toodle with the new Compliance Documentation page. They are not in the Braze org, so their ability to collaborate is limited. I offered to make it a hidden doc so they could fork it and pull the changes in that way. 

@bre-fitzgerald , let me know if you have any concerns about this methodology.

bre-fitzgerald (approved):

PR #6726, opened 2024-01-12
BD-2726 Add Custom Attributes tile to Segments page

rachel-feinberg:

Closes #BD-2726

- [x] No

### DIFF
--- a/_docs/_user_guide/engagement_tools/segments.md
+++ b/_docs/_user_guide/engagement_tools/segments.md
@@ -52,4 +52,11 @@ guide_menu_list:
   - name: Regular Expressions
     link: /docs/user_guide/engagement_tools/segments/regex/
     fa_icon: fas fa-search
+
+guide_menu_title: "Other"
+guide_menu_list:
+  - name: Custom Attributes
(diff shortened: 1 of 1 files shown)


_docs/_user_guide/engagement_tools/segments.md
```diff
@@ -52,4 +52,11 @@ guide_menu_list:
   - name: Regular Expressions
     link: /docs/user_guide/engagement_tools/segments/regex/
     fa_icon: fas fa-search
+
+guide_menu_title: "Other"
+guide_menu_list:
```
bre-fitzgerald, line 57: When we add more sections, we have to increment a counter; otherwise it will override the previous sections. If you check the deployment rn, it's only showing "Other" because there are two `guide_menu_title`s in the file, and this is the one at the end.
```suggestion
guide_menu_title2: "Other"
guide_menu_list2:
```

rachel-feinberg: Thanks for catching this! It didn't register 🙇‍♀️

bre-fitzgerald (approved):

PR #6736, opened 2024-01-16
BD-2741: New help article for reducing push notification payload size

bre-fitzgerald:

Closes #**BD-2741**

- [ ] No

---

### DIFF
--- a/_docs/_help/help_articles/push/reducing_payload_size.md
+++ b/_docs/_help/help_articles/push/reducing_payload_size.md
@@ -23,13 +23,13 @@ Push service providers calculate whether your push notification can be displayed
 
 These push services will refuse any notification that exceeds this limit.
 
-Braze reserves a portion of the push payload for integration and analytics purposes. Given that, our recommended payload size is **3,807 bytes**. If your push exceeds this size, part of your message may be cut off, or the push may not be sent at all.
+Braze reserves a portion of the push payload for integration and analytics purposes. Given that, our maximum payload size is **3,807 bytes**. If your push exceeds this size, the message may not be sent. As a best practice, keep your payload to a few hundred bytes.
 
 The following elements in your push make up your push payload:
 
 - Copy, such as the title and message body
 - Final render of any Liquid personalization
-- Images
+- URLs for images (but not the size of the image itself)
 - URLs for click targets
 - Button names
 - Key-value pairs
--- a/_docs/_help/help_articles/push/reducing_payload_size.md
+++ b/_docs/_help/help_articles/push/reducing_payload_size.md
@@ -10,7 +10,7 @@ channel: push
 
 If you can't launch a push campaign or Canvas step because your push payload is too large, check out these tips for reducing your push notification payload size.
 
-> Our recommended payload size is **3,807 bytes**. If your push exceeds this size, part of your message may be cut off, or the push may not be sent at all.
+> Our maximum payload size is **3,807 bytes**. If your push exceeds this size, the message may not be sent. As a best practice, keep your payload to a few hundred bytes.
 
 ## What is a push payload?
 
--- a/_docs/_help/help_articles/push/reducing_payload_size.md
+++ b/_docs/_help/help_articles/push/reducing_payload_size.md
@@ -1,0 +1,44 @@
+---
+nav_title: Reducing Push Notification Payload Size
+article_title: Reducing Push Notification Payload Size
+page_type: solution
+description: "This help article provides some tips to reduce the payload size of your push notifications if you're unable to launch a campaign or Canvas step due to push payload size limits."
+channel: push
+---
+
+# Reducing push notification payload size
+
+If you can't launch a push campaign or Canvas step because your push payload is too large, check out these tips for reducing your push notification payload size.
+
+> Our recommended payload size is **3,807 bytes**. If your push exceeds this size, part of your message may be cut off, or the push may not be sent at all.
+
+## What is a push payload?
+
+Push service providers calculate whether your push notification can be displayed to a user by looking at the byte size of the entire push payload. The payload is limited to **4KB (4,096 bytes)** for most push services, including:
+
+- Apple Push Notification service (APNs)
+- Android’s Firebase Cloud Messaging (FCM)
+- Web push
+- Huawei push max
+
+These push services will refuse any notification that exceeds this limit.
+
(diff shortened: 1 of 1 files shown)


_docs/_help/help_articles/push/reducing_payload_size.md
```diff
@@ -0,0 +1,44 @@
...
+page_type: solution
+description: "This help article provides some tips to reduce the payload size of your push notifications if you're unable to launch a campaign or Canvas step due to push payload size limits."
+channel: push
+---
+
+# Reducing push notification payload size
+
+If you can't launch a push campaign or Canvas step because your push payload is too large, check out these tips for reducing your push notification payload size.
+
+> Our recommended payload size is **3,807 bytes**. If your push exceeds this size, part of your message may be cut off, or the push may not be sent at all.
```
froodian, line 13: I don't think there's ever really a situation due to byte size where part of a message would get cut off... there are UI situations where visibly the OS cuts them off, but at much smaller sizes than these limits (those should generally get reflected in the composer previews, though it's not perfect given different device sizes and aspect ratios).  

It also makes me realize, pedantically, I would _recommend_ payloads of like a few hundred bytes, just because short messages are best suited for push from a use-case perspective (eg to keep anything from being visually cut off) - these max sizes are more like upper limits than recommended sizes.

_docs/_help/help_articles/push/reducing_payload_size.md
```diff
@@ -0,0 +1,44 @@
...
+
+## What is a push payload?
+
+Push service providers calculate whether your push notification can be displayed to a user by looking at the byte size of the entire push payload. The payload is limited to **4KB (4,096 bytes)** for most push services, including:
+
+- Apple Push Notification service (APNs)
+- Android’s Firebase Cloud Messaging (FCM)
+- Web push
+- Huawei push max
```
froodian, line 22:
```suggestion
- Huawei push
```

_docs/_help/help_articles/push/reducing_payload_size.md
```diff
@@ -0,0 +1,44 @@
...
+
+These push services will refuse any notification that exceeds this limit.
+
+Braze reserves a portion of the push payload for integration and analytics purposes. Given that, our recommended payload size is **3,807 bytes**. If your push exceeds this size, part of your message may be cut off, or the push may not be sent at all.
+
+The following elements in your push make up your push payload:
+
+- Copy, such as the title and message body
+- Final render of any Liquid personalization
+- Images
```
froodian, line 32: i'm realizing that I think for all these channels, I think it's just the URLs of the images that contributes to the payload's bytesize

froodian (approved):

_docs/_help/help_articles/push/reducing_payload_size.md
bre-fitzgerald: @froodian good catch, thanks! would you say it's better/more accurate to say something like this?

```suggestion
> Our maximum payload size is **3,807 bytes**. If your push exceeds this size, the message may not be sent. As a best practice, keep your payload to a few hundred bytes.
```

_docs/_help/help_articles/push/reducing_payload_size.md
bre-fitzgerald: The content doesn't have an impact? If not, why would Liquid personalization be such a factor?

_docs/_help/help_articles/push/reducing_payload_size.md
froodian: looks good to me!

_docs/_help/help_articles/push/reducing_payload_size.md
froodian: The content of the push has impact - and the image url is part of that, and can be a thing that contains liquid - but the image itself just isn't part of the push, just the url is sent in the push and then the device downloads the image bytes itself from that URL, so the size of the image itself is exempted from the push payload limit.

In terms of why liquid is SUCH a factor, I think it's mostly a big deal in the title and message body where most language localization is done with liquid, so even if the title is just "Hi" that might be 20 liquid if/else conditions for each language, followed by "Hi" in each language

bre-fitzgerald: Ohhh my mistake, I misunderstood your original comment here 🤦‍♀️ Will update that last bullet accordingly

_docs/_help/help_articles/push/reducing_payload_size.md
```diff
@@ -0,0 +1,44 @@
...
+page_type: solution
+description: "This help article provides some tips to reduce the payload size of your push notifications if you're unable to launch a campaign or Canvas step due to push payload size limits."
+channel: push
+---
+
+# Reducing push notification payload size
+
+If you can't launch a push campaign or Canvas step because your push payload is too large, check out these tips for reducing your push notification payload size.
+
+> Our maximum payload size is **3,807 bytes**. If your push exceeds this size, the message may not be sent. As a best practice, keep your payload to a few hundred bytes.
```
lydia-xie, line 13:
```suggestion
> Our maximum payload size is **3,807 bytes**. If your push exceeds this size, the message may not be sent. As a best practice, keep your payload to a few hundred bytes.
```
Thoughts on making this a tip or note instead?

lydia-xie (approved):
lgtm! just left 1 small suggestion/comment :)

_docs/_help/help_articles/push/reducing_payload_size.md
bre-fitzgerald: Good question! I was using it as the actual usage for pull-out quotes (this content is repeated farther down in context, but I wanted to make sure it was very, very explicitly stated). Is this confusing because we use block quote syntax for top-of-page descriptions?

lydia-xie: > Is this confusing because we use block quote syntax for top-of-page descriptions?

Yeah, that was my thinking! Don't know of any other instances where we use block quote syntax for something like this

bre-fitzgerald: Yeah, I've done it in at least one other place, but I can't recall exactly which article. I'll just change it to a note. 👍

bre-fitzgerald, line 13:
```suggestion
{% alert note %}
Our maximum payload size is **3,807 bytes**. If your push exceeds this size, the message may not be sent. As a best practice, keep your payload to a few hundred bytes.
{% endalert %}
```

PR #6819, opened 2024-01-31
Removed code styling

josh-mccrowell-braze:
The external ID parameter naming convention changes between the SDK paradigm (`external_id`) and the API (`external_user_id`). To thread the needle, I removed the code styling and just talked about user IDs.


internetisaiah (approved):
LGTM

PR #6831, opened 2024-02-05
Update message_extras.md

davidbielik:


_docs/_user_guide/personalization_and_dynamic_content/liquid/advanced_filters/message_extras.md
```diff
@@ -54,7 +54,7 @@ A `dispatch_id` is generated and provided in your send events, which can be used
 
 #### Can I use message_extras with in-app messages?
 
-The `message_extras` tag is supported for all message types with a send event. In-app messages do not have send events, so `message_extras` are not available for in-app messages.
+Support for `message_extras` in the in-app message channel is coming soon!
```
bre-fitzgerald, line 57: @davidbielik per our style guide, we avoid referring to future features in documentation:

> Avoid references to future features, or suggestions that something may be supported in the future.
> Don’t use words and phrases that anchor your writing to a point in time, as they make content become quickly outdated. Focus on how the product works right now, not on what has changed (except for time-focused content, such as in release notes).

Generally I would suggest not making this change until `message_extras` are supported, but if we must, I would suggest language like this:

```suggestion
Using`message_extras` in the in-app message channel is not supported at this time.
```

cc: @josh-mccrowell-braze @internetisaiah

davidbielik: it says "avoid" does that mean never? even if we are actively building it?

internetisaiah: That's a good note @davidbielik , and something maybe we can consider slightly rephrasing in the style guide to be more clear. Thanks!

But to answer your question, I think a general rule of thumb is that content should "sound like docs" instead of "sound like a blog." If you'd like to enhance Bre's update to add some more specifics, you could do something like:

```
Using`message_extras` in the in-app message channel is not supported at this time. For more information, see [Braze February Updates](www.example.com/some/random/blog).
```

internetisaiah (approved):
LGTM

internetisaiah:
@davidbielik , if you're fine with Bre's edits i'll merge. Otherwise, let me know if you want any tweaks before I merge 👍🏽

PR #7145, opened 2024-04-02
Lowercase folder names

josh-mccrowell-braze:
You can't tell, but I made the folder names lowercase for the Shopify docs


rachel-feinberg (approved):
Lgtm! @josh-mccrowell-braze, will this impact the other redirects and links that use the "Shopify" path?

PR #7211, opened 2024-04-12
Building out some small changes to the Canvas documentation

josh-mccrowell-braze:


josh-mccrowell-braze:
Fixes applied in https://github.com/braze-inc/braze-docs/pull/7224

PR #7290, opened 2024-05-01
Basic wireframe for a new tab type

zzhaobraze:

## Usage:

```
{% sdktabs %}
{% sdktab tab1 %}
Tab1 Content
{% endsdktab %}

{% sdktab Tab2 %}
Tab2 Content
{% endsdktab %}
{% endsdktabs %}
```

## Example:

image

### DIFF
--- a/_layouts/broken_page.html
+++ b/_layouts/broken_page.html
@@ -81,7 +81,8 @@ if (site_language && (site_language != 'en')){
   if (redirected_count < max_redirects) {
     redirected_count++
     query_params.set('redirected',redirected_count );
-    var redirected  = '?' + query_params.toString();
+
+    var redirected  = ((urlsearch.indexOf('?') > -1 ) ? '?' : '&') + query_params.toString();
     if (validurls[urlpath.toLowerCase() + '/' + urlhash] ) {
       return location.href = redirecturl(urlpath.toLowerCase() + '/' + urlhash,urlhash,redirected);
     }
--- a/_plugins/sdktabs.rb
+++ b/_plugins/sdktabs.rb
@@ -1,38 +1,39 @@
 require 'digest/md5'
 
 module Tags
-    class TabsBlock < Liquid::Block
+    class SdkTabsBlock < Liquid::Block
       def initialize(tag_name, tabonly = 'false', tokens)
           super
-          @tabclass = 'tab_toggle'
-          @tabid = 'tab_' + (0...12).map { (97 + rand(26)).chr }.join
+          @tabclass = 'sdk-tab_toggle'
+          @tabid = 'sdk-tab_' + (0...12).map { (97 + rand(26)).chr }.join
           if tabonly.downcase.strip == 'local'
-            @tabclass = 'tab_toggle_only'
+            @tabclass = 'sdk-tab_toggle_only'
           end
       end
       def render(context)
-          tabs = super.scan(/data\-tab=\"(.*?)\"/)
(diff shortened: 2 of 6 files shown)


internetisaiah:
Notes:

1. The tab should still show even if there is only 1 `{% sdktabs %}` on the page. This is the default behavior on the existing `{% tabs %}` tag.
2. The new SDK tab is reading both `{% sdktabs %}` _and_ `{% tabs %}` as SDK tabs. Instead, regular tabs should be able to be fully nested inside SDK tabs. (See image below for existing behavior).
3. When an SDK tab is selected, the TOC should only show the headings for the currently selected SDK tab. Currently all headings are shown regardless of the SDK tab selected.

image

cc @zzhaobraze

zzhaobraze:
@internetisaiah the previous commit should fix this issue. If you update the branch and resolve the conflict, this should work now.

internetisaiah:
@zzhaobraze , thanks for the update! i already merged it into my other branch on friday and have been testing it out. works great!

internetisaiah:
@zzhaobraze , similar to `{% tabs %}` and `{% subtabs %}`, could we get an `{% sdktabs %}` and `{% sdksubtabs %}`?

As far as design goes, it can look and function exactly like a regular tab.

However in terms of nesting, it should go:

- Highest level: `{% sdktabs %}`
- Next level: `{% sdksubtabs %}`
- Next level: `{% tabs %}`
- Lowest level: `{% subtabs %}`

internetisaiah:
ok @zzhaobraze , everything is looking amazing! 🙌🏽  thanks for getting it to the state it's in currently. i've only found one last "bug" that's a priority to fix. all the other bugs are low impact and can be tackled down the road though.

when setting up a redirect like this:

```js
validurls['/docs/developer_guide/platforms/web/sdk_integration/'] = '/docs/developer_guide/sdk_integration/?sdktab=web';
``` 

the redirect works, but it always selects the android tab by default (regardless of the URL ending). not sure if this helps troubleshoot, but this is the URL in the browser after the direct. not sure if its like an order of operations thing?

```
http://127.0.0.1:5006/docs/developer_guide/sdk_integration/?sdktab=swift?redirected=1
```

here's a screenshot of what i see in the browser:

Screen Shot 2025-02-10 at 17 51 12 PM

internetisaiah:
@zzhaobraze , also, i updated the colors in [this commit](https://github.com/braze-inc/braze-docs/pull/7290/commits/639bf531d98a9df1c6a4f8bcb019fbdebe4acd43) to better match the standard tab colors (the previous colors didn't have as much contrast so i wanted to make them easier to read. let me know if i changed any of the wrong lines--i'm happy to fix it 👍🏽

zzhaobraze:
> @zzhaobraze , also, i updated the colors in [this commit](https://github.com/braze-inc/braze-docs/pull/7290/commits/639bf531d98a9df1c6a4f8bcb019fbdebe4acd43) to better match the standard tab colors (the previous colors didn't have as much contrast so i wanted to make them easier to read. let me know if i changed any of the wrong lines--i'm happy to fix it 👍🏽

i fix the redirect issue. was a bug due to the multi-redirect change. it should be an '&' instead of a '?'
i would make the contrast more obvious, let me know if you need help with that.

internetisaiah:
> If there is a question of contrast, can we loop Jucelle in to the conversation to make sure we're picking ADA compliant colors?

for now i've set the colors to match our existing tabs 1:1 (which i'm assuming already meet this standard), but i do plan on looping in a designer in the next few weeks to see if we can get some tweaks to the design (but that would be separate work that doesn't impact this PR from being merged).

PR #7350, opened 2024-05-15
Update disabling_tracking.md

davidbielik:


_docs/_developer_guide/platform_integration_guides/swift/analytics/disabling_tracking.md
```diff
@@ -15,4 +15,6 @@ When set to `false`, the Braze SDK ignores any call to the public API. The SDK a
...
+Prior to Swift SDK v5.7.0, IDFV was used as the default device ID. On these older SDK versions, or when setting `useUUIDAsDeviceId` to `false`, additional steps may be required to delete device information. In these cases, unless a user uninstalls all apps from a vendor on a given device, the next time the Braze SDK runs after calling `wipeData()` will result in our server re-identifying that user via their device identifier. In order to fully remove all user data, you should combine a call to `wipeData()` with a request to delete data on the server via the Braze [REST API]({{site.baseurl}}/api/endpoints/user_data/post_user_delete/).
+
+Swift SDK versions 5.7.0 and above default to using a randomly generated UUID as the device ID, and calling `wipeData()` will result in a new device ID to be randomly generated.
```
jerielng, line 20: This only became the default in `7.0.0`. The configuration was made available in `5.7.0`, but we were still defaulting to IDFV.
```suggestion
Swift SDK versions 7.0.0 and above default to using a randomly generated UUID as the device ID, and calling `wipeData()` will result in a new device ID to be randomly generated.
```

internetisaiah (approved):
LGTM

PR #7466, opened 2024-06-05
File content_management.md committed.

Translationmanagement:


josh-mccrowell-braze:
@zzhaobraze Just checking in on this. Are we still holding on this and https://github.com/braze-inc/braze-docs/pull/7467?

zzhaobraze:
i believe this might be a test, @Translationmanagement (iara) should confirm

PR #7801, opened 2024-08-05
Fix punctuation issue

zzhaobraze:
Fix issue with punctuation error causing js error on the page.


josh-mccrowell-braze:
@cla-bot check

PR #7960, opened 2024-08-26
[FF-588] - Update web_objects.md

wesleyorbin:
### Summary

This updates the description of the `require_interaction` field of the web push object to reflect the current support landscape. It is based on the MDN documentation [here](https://developer.mozilla.org/en-US/docs/Web/API/Notification/requireInteraction#browser_compatibility).

### Related PRs, issues, or features (optional)

https://jira.braze.com/browse/FF-588

### DIFF
--- a/_docs/_api/objects_filters/messaging/web_objects.md
+++ b/_docs/_api/objects_filters/messaging/web_objects.md
@@ -23,7 +23,7 @@ description: "This reference article lists and explains the different web object
    "custom_uri": (optional, string) a web URL,
    "image_url": (optional, string) URL for image to show,
    "large_image_url": (optional, string) URL for large image, supported on Chrome Windows/Android,
-   "require_interaction": (optional, boolean) whether to require the user to dismiss the notification, supported on Mac Chrome,
+   "require_interaction": (optional, boolean) whether to require the user to dismiss the notification, supported on Firefox for Windows, Chrome, Edge, and Opera,
    "time_to_live": (optional, integer (seconds)),
    "send_to_most_recent_device_only" : (optional, boolean) defaults to false, if set to true, Braze will only send this push to a user's most recently used browser, rather than all eligibles browsers,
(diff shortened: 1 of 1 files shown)


_docs/_api/objects_filters/messaging/web_objects.md
```diff
@@ -23,7 +23,7 @@ description: "This reference article lists and explains the different web object
    "custom_uri": (optional, string) a web URL,
    "image_url": (optional, string) URL for image to show,
    "large_image_url": (optional, string) URL for large image, supported on Chrome Windows/Android,
-   "require_interaction": (optional, boolean) whether to require the user to dismiss the notification, supported on Mac Chrome,
+   "require_interaction": (optional, boolean) whether to require the user to dismiss the notification, supported on Firefox for Windows, Chrome, Edge, and Opera,
```
wesleyorbin, line 26: Is it possible to link to the MDN documentation from within this code block?

internetisaiah:
```suggestion
   "require_interaction": (optional, boolean) whether to require the user to dismiss the notification. for list of supported platforms, see: "https://developer.mozilla.org/en-US/docs/Web/API/Notification/requireInteraction#browser_compatibility",
```

Would something like this work? Alternatively, I can see if its against our style guide to use short URLs? So at least it won't be as long?

That being said, there's no way to do this `[MDN docs](url_string)` _within_ a codeblock on Braze Docs.

internetisaiah: @wesleyorbin , just tagging you here in case the notification fell through the cracks. Let me know if my suggestion above works for you 👍🏽

wesleyorbin: That suggestion works for me. Thank you!

internetisaiah (approved):
LGTM

PR #8000, opened 2024-08-30
Update browser_extensions.md

davidbielik:


_docs/_developer_guide/platform_integration_guides/web/browser_extensions.md
```diff
@@ -25,7 +25,7 @@ In general, since extensions are HTML and JavaScript, you can use Braze for the
 
 ## What's not supported
 
-* Manifest v3 service workers do not support importing modules meant for web environments.
+* Service workers for Manifest v3 is not supported due to lack of DOM/Window API support. This is on our roadmap for future consideration.
```
internetisaiah, line 28:
```suggestion
* Service workers for Manifest v3 are not supported due to a lack of a Document Object Model (DOM) and Window API support. However, this is on the roadmap for future consideration.
```

internetisaiah (approved):
LGTM

_docs/_developer_guide/platform_integration_guides/web/browser_extensions.md
davidbielik: @internetisaiah on second thought maybe we just explain that our SDK is not supported in this env.

PR #8607, opened 2024-12-10
Update faq.md

SimeonBrazeSA:
It is now possible to create segments of users that have a feature flag I removed this


josh-mccrowell-braze:
Thanks for flagging, @SimeonBrazeSA!

@bkasman95 This is a good example of the sort of forward looking statement that grows stale. I think, in general, we should avoid these.

bkasman95 (approved):

PR #8621, opened 2024-12-12
clean up fr dir and remove deprecated banner

zzhaobraze:
Remove old french dir, and deprecated banner


josh-mccrowell-braze (approved):
Clicked around for a while. Looks OK to me. Waiting for Iara's review, then will merge.

josh-mccrowell-braze (approved):
Going ahead and merging so it goes out today

PR #8653, opened 2024-12-18
auto expand current nav if possible

zzhaobraze:
Auto expand if current page navigation is expandable
image


internetisaiah (approved):
LGTM! thanks for the quick turn around @zzhaobraze 🫡

PR #8755, opened 2025-01-13
make external links open its own tab

zzhaobraze:
### Why are you making this change? (required)
Make external link open a new tab

### DIFF
--- a/assets/js/documents.js
+++ b/assets/js/documents.js
@@ -575,6 +575,7 @@ $(document).ready(function() {
     }
     if (is_external){
       $(this).after(' <i class="fas fa-external-link-alt"></i>');
+      $(this).attr('target', '_blank')
     }
   });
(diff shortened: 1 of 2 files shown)


assets/js/documents.js
```diff
@@ -575,6 +575,7 @@ $(document).ready(function() {
     }
     if (is_external){
       $(this).after(' <i class="fas fa-external-link-alt"></i>');
+      $(this).attr('target', '_blank')
```
bre-fitzgerald, line 578: Hey @zzhaobraze! It looks like the deployment failed for this PR. Is this line perhaps missing an ending semicolon? `;`

zzhaobraze: the ; is optional, but this is due to the deployment issue. i can retrigger by adding the ;

bre-fitzgerald (approved):

PR #8976, opened 2025-02-11
fix redirects

zzhaobraze:
Fix redirects `?` being replace incorrectly


internetisaiah (approved):
LGTM, it works for regular redirects and sdktab redirects.

PR #9010, opened 2025-02-18
fix missing partner

zzhaobraze:
Partner page seems to generate an error when a partner is missing with the new schema.


josh-mccrowell-braze (approved):
Partner page and landing pages now load appropriately

PR #9043, opened 2025-02-21
[CO-3331] Clarifies data point consumption

shihanxiong:
https://jira.braze.com/browse/CO-3331


LiZhangBraze (approved):
did we want to add a note about how blocklisted data doesn't count?

shihanxiong:
> did we want to add a note about how blocklisted data doesn't count?

I think this was already mentioned in https://www.braze.com/docs/user_guide/data/data_points/#viewing-data-point-usage

lydia-xie (approved):

PR #9066, opened 2025-02-25
CO-3336 add note about aliases with the same label

mkigikm:
[CO-3336](https://jira.braze.com/browse/CO-3336)

Add a note about what happens if two users in an identify scenario both have an alias with the same label.


lydia-xie (approved):
lgtm!

LiZhangBraze (approved):

PR #9082, opened 2025-02-26
CO-3355: update user profile lifecycle page to clarify a static braze_id + add email and phone identifiers

LiZhangBraze:
[CO-3355](https://jira.braze.com/browse/CO-3355)
- add a line to clarify that braze_id is set by braze (cannot be assigned)
- include email and phone as identifiers


lydia-xie (approved):
lgtm!

PR #9175, opened 2025-03-07
Fixed broken double slash links

davidbielik:
regex: `(?<!^)(?<![\/#:\\: ])//`


josh-mccrowell-braze:
Oh, great idea, David. Thank you for putting this in. Is this one ready to go?

PR #9176, opened 2025-03-07
Script to find broken links.

davidbielik:
1. `yarn install`

2. `npx ts-node get_links.ts`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,5 @@ braze-docs.iml
 
 # Braze scripts
 scripts/temp/*
+node_modules
+broken-links.csv
--- a/assets/js/broken_redirect_list.js
+++ b/assets/js/broken_redirect_list.js
@@ -2039,3 +2039,9 @@ validurls['/docs/developer_guide/platforms/swift/initialization/delayed_initiali
 validurls['/docs/developer_guide/platforms/unreal_engine/initialization/'] = '/docs/developer_guide/sdk_initalization/?sdktab=unreal%20engine';
 
 // validurls['OLD'] = 'NEW';
+
+if (typeof module !== 'undefined' && module.exports) {
+  // export so i can import validurls from another file
+  module.exports = validurls;
+
+}
--- a/get_links.ts
+++ b/get_links.ts
@@ -1,0 +1,244 @@
+import { existsSync, link } from 'fs';
+import path, { resolve } from 'path';
+import fs from 'fs';
+
+const redirectList: string[] = Object.keys(require('./assets/js/broken_redirect_list.js'));
+
+interface LinkData {
+  sourceFile: string;
+  link: string;
+  markdownFile: string;
+}
+
+/**
+ * Converts a URL path to its corresponding markdown file path
+ * @param link - The URL path to convert (e.g., "/user_guide/administrative")
+ * @returns The corresponding markdown file path
+ */
+const convertLinkToMarkdownPath = (link: string): string => {
+  // Remove leading and trailing slashes
+  const trimmedLink = link.replace(/^\/|\/$/g, '');
+
+  // Split the path into segments
+  const segments = trimmedLink.split('/');
+
+  // If there's only one segment, return simple _docs path
+  if (segments.length === 1) {
+    return `_docs/${segments[0]}.md`;
+  }
+
+  // Add _docs prefix and underscore to first segment
+  const docPath = `_docs/_${segments[0]}`;
+
+  // Add remaining path segments
+  const remainingPath = segments.slice(1).join('/');
+
+  // Combine paths and add .md extension
+  return `${docPath}/${remainingPath}.md`;
+};
+
+const getLinks = (filePath: string): {links: LinkData[], aliases: string[], permalinks: string[]} => {
+  if (!filePath) {
+    process.exit(1);
+  }
+
+  let content: string;
+  try {
+    content = fs.readFileSync(filePath, 'utf8');
+  } catch (err) {
+    console.error(`Error reading file: ${err}`);
+    process.exit(1);
+  }
+
+  /**
+   * Remove templated base URL, trim whitespace, and remove any hash fragments.
+   */
+  function cleanUrl(url: string): string {
+    // Remove {{ site.baseurl }} variations (with or without spaces)
+    url = url.replace(/{{\s*site\.baseurl\s*}}/g, '');
+    url = url.trim();
+    // Remove any hash fragment (e.g., /with/path#hash becomes /with/path)
+    url = url.split('#')[0].trim();
+    url = url.split('?')[0].trim();
+    url = url.split(' ')[0].trim();
+    return url;
+  }
+
+  /**
+   * Returns true if the URL:
+   * - Starts with a slash (i.e. a full/relative URL)
+   * - And its last segment does not look like a file (i.e. has no extension).
+   */
+  function isValidLink(url: string): boolean {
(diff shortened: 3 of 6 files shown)


assets/js/broken_redirect_list.js
```diff
@@ -2042,3 +2042,9 @@ validurls['/docs/developer_guide/platforms/unreal_engine/initialization/'] = '/d
 validurls['/docs/developer_guide/sdk_initalization/'] = '/docs/developer_guide/sdk_initialization/';
 
 // validurls['OLD'] = 'NEW';
+
+if (typeof module !== 'undefined' && module.exports) {
```
internetisaiah, line 2046: is this able to be moved to the top of the page? or does it need to be explicitly be called after all the `validurls` lists?

scripts/package-lock.json
```diff
@@ -0,0 +1,27 @@
+{
```
internetisaiah, line 1: i believe all the `.json` files for packages/configs can be moved to root. then just keep the actual `find_broken_links.ts` script in `./scripts`.

internetisaiah (commented):
will do further testing, but left a couple notes on some files 👍🏽

assets/js/broken_redirect_list.js
davidbielik: needs to be at the end to export the variable that's been defined.

scripts/package-lock.json
davidbielik: didn't want to muddy up the root directory in case in the future you use npm but up to you

assets/js/broken_redirect_list.js
internetisaiah: got it. good to know. just wanted to verify 👍🏽

.gitignore
```diff
@@ -11,12 +11,12 @@ jekyll-algolia-*
 .jekyll-cache
 .vscode
 .jekyll-metadata
+yarn.lock
```
internetisaiah, line 14: been a while since i've used yarn. should we be adding the `yarn.lock` here or naw? @davidbielik

bdocs
```diff
@@ -122,6 +123,10 @@ case $1 in
         "$ULINKS" "$2"
         # rm "$REDIRECT_MATCHES"
         ;;
+    fblinks)
```
internetisaiah, line 126: note: added this to our wrapper script `bdocs`. you can still run the command directly like this:

```
npx ts-node ./scripts/find_broken_links.ts
```

but you can optionally do this:

```
./bdocs fblinks
```

scripts/find_broken_links.ts
```diff
@@ -0,0 +1,252 @@
...
+ *
+ * Usage: ./bdocs fblinks
+ */
+
+import { existsSync, link } from 'fs';
+import path, { resolve } from 'path';
+import fs from 'fs';
+import { execSync } from 'child_process';
+
+const PROJECT_ROOT = execSync('git rev-parse --show-toplevel', { encoding: 'utf8' }).trim();
```
internetisaiah, line 13: note: created a new var for the project root. since this will always be run in a git env, this set's project root to git root so you can run this script from anywhere.

scripts/find_broken_links.ts
```diff
@@ -0,0 +1,252 @@
...
+
+const redirectList: string[] = Object.keys(require(path.join(PROJECT_ROOT, 'assets/js/broken_redirect_list.js')));
```
internetisaiah, line 15: @davidbielik , i know nothing about typescript syntax so asked chatgpt for help. seems to be similar to python's `path.join()` but let me know if there's a better approach.

internetisaiah (changes requested):
@davidbielik , left some notes throughout for you to double check. no changes to be made for them. just for verifying changes i made in the last few commits.

additionally, is there anyway to only print `false` links by default? then maybe pass in some flag like `-v` or `--verbose` to print `flase` _and_ `true`? from my testing, the true prints makes it hard to find the false one in the list.

EDIT: _unless_ true means that broken == true, then in that case, i'll just add that note to our docs and you can ignore.

scripts/find_broken_links.ts
```diff
@@ -0,0 +1,252 @@
...
+  //   continue;
+  // }
+  // if (!result.links) {
+  //   console.log(`No links found`);
+  //   console.log(result);
+  //   continue;
+  // }
+
+  // For each link in the result
+  // Add to CSV if link exists is false
```
internetisaiah, line 218: @davidbielik , for "Expected Path", just want to confirm the meaning. if csv is:

```csv
/Users/Isaiah.Robinson/braze-docs/_docs/_api/api_limits.md,/docs/api/endpoints/email/bounce/remove,/Users/Isaiah.Robinson/braze-docs/_docs/_api/endpoints/email/bounce/remove.md
```

that means that the file that's getting flagged is:
```
/Users/Isaiah.Robinson/braze-docs/_docs/_api/api_limits.md
```

the broken URL is:
```
/docs/api/endpoints/email/bounce/remove
```

and it should be set to:
```
/docs/api/endpoints/email/bounce/remove
```

.gitignore
davidbielik: No, that shouldn't be there - good catch.

davidbielik:
```suggestion
```

scripts/find_broken_links.ts
```diff
@@ -0,0 +1,252 @@
...
+      links.push(...data.links);
+      aliases.push(...data.aliases);
+      permalinks.push(...data.permalinks);
+      totalFiles++;
+    }
+  }
+}
+
+getLinksRecursive(docsBasePath);
+console.log(`Found ${links.length} links`);
```
davidbielik, line 198:
```suggestion
```
shouldn't have been here in the first place

scripts/find_broken_links.ts
davidbielik: it expected to find `/_docs/_api/endpoints/email/bounce/remove.md`

internetisaiah: hmm, isn't the broken link and the file it expected to find the same in this case? or am i understanding one of these differently?

davidbielik: in your example, the `/api_limits` page linked out to the `/remove` page which should have been found at `/remove.md` file

internetisaiah: is there ever a case were column 2 and 3 are not the same? (just one is the URL version vs the path version)

scripts/find_broken_links.ts
```diff
@@ -242,4 +249,4 @@ const deduplicated = Array.from(new Set(csv)).sort();
 
 console.log(`\n\nFound ${deduplicated.length - 1} broken links in ${totalFiles} files`);
 // console.log(redirectList)
-fs.writeFileSync('broken-links.csv', [headers, ...deduplicated].join('\n'));
+fs.writeFileSync('./scripts/temp/broken-links.csv', [headers, ...deduplicated].join('\n'));
```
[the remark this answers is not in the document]
internetisaiah: note, i switched it to just always use the `PROJECT_ROOT` var so it can be run in all these cases:

```bash
# case 1
./bdocs fblinks

# case 2
npx ts-node ./scripts/find_broken_links.ts 

# case 3
cd scripts
npx ts-node ./find_broken_links.ts 
```

internetisaiah (approved):
LGTM. thanks for your work on this! gonna be _super_ helpful!

PR #9381, opened 2025-04-04
Docs Developer Profile

davidbielik:
implements https://github.com/Appboy/platform/pull/95795 to set developer profile in Braze SDK and changes to new Braze Workspace.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -11,7 +11,7 @@ github_username: braze-inc
 language: en
 
 # Braze Web SDK Key
-api_key: "5b26b83b-aa4e-4bd1-b9fa-81880aa8ac83"
+api_key: "c574d66e-2a74-4ef2-b2f0-ae004a66f682"
 
 # statuspage.io Key
 status_io_key: "c947v6z63ltm"
--- a/assets/developer-sync.html
+++ b/assets/developer-sync.html
@@ -1,0 +1,21 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Developer Sync</title>
+  </head>
+  <body>
+    <script type="text/javascript">
+      // Listen for postMessage events from Braze domains
+    window.addEventListener('message', function(event) {
+        const isAllowedOrigin = /^https:\/\/.*\.braze\.(com|eu)$/.test(event.origin) || /^http:\/\/localhost:/.test(event.origin);
+        // Check if the origin is from a braze.com or braze.eu domain
+        if (isAllowedOrigin) {
+        // Process the message
+        if (event.data && event.data?.type === 'BRAZE_DEVELOPER_DATA') {
+            window.localStorage.setItem('braze_developer_data', JSON.stringify(event.data.payload));
+        }
+      }
+    });
+    </script>
+  </body>
+</html>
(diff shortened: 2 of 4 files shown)


_config.yml
```diff
@@ -11,7 +11,7 @@ github_username: braze-inc
 language: en
 
 # Braze Web SDK Key
-api_key: "5b26b83b-aa4e-4bd1-b9fa-81880aa8ac83"
+api_key: "c574d66e-2a74-4ef2-b2f0-ae004a66f682"
```
davidbielik, line 14: waiting to confirm braze@braze campaigns will need to move to this workspace.

davidbielik: Laina confirmed this is good to go and we don't care about the content cards anymore. Just have to recreate the SQL Query for 404 pages.

assets/developer-sync.html
```diff
@@ -0,0 +1,21 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Developer Sync</title>
+  </head>
+  <body>
+    <script type="text/javascript">
+      // Listen for postMessage events from Braze domains
+    window.addEventListener('message', function(event) {
```
zzhaobraze, line 9: this is probably better done without using test
(double check this works)
```suggestion
    window.addEventListener('message', (event) => {
  try {
    const url = new URL(event.origin);
    const { protocol, hostname } = url;
  
    // Check if it's from localhost (http only)
    const isLocalhost = protocol === 'http:' && hostname.startsWith('localhost');
  
    // Check if it's from a Braze domain (must be https)
    const isBrazeDomain =
      protocol === 'https:' &&
      (hostname.endsWith('.braze.com') || hostname.endsWith('.braze.eu'));
  
    if (!isLocalhost && !isBrazeDomain) return;
  
    // Process the message
    const { type, payload } = event.data || {};
    if (type === 'BRAZE_DEVELOPER_DATA' && payload) {
      window.localStorage.setItem('braze_developer_data', JSON.stringify(payload));
    }
  } catch (e) {
    // event.origin might not be a valid URL
    console.error('Invalid origin:', event.origin);
  }
});
```

assets/developer-sync.html
```diff
@@ -0,0 +1,21 @@
...
+        const isAllowedOrigin = /^https:\/\/.*\.braze\.(com|eu)$/.test(event.origin) || /^http:\/\/localhost:/.test(event.origin);
+        // Check if the origin is from a braze.com or braze.eu domain
+        if (isAllowedOrigin) {
+          // Process the message
+          if (event.data && event.data?.type === 'BRAZE_DEVELOPER_DATA') {
+              window.localStorage.setItem('braze_developer_data', JSON.stringify(event.data.payload));
+          }
+        }
+    });
```
zzhaobraze, line 18: this is probably better done without using test
(double check this works)

```suggestion
   window.addEventListener('message', (event) => {
  try {
    const url = new URL(event.origin);
    const { protocol, hostname } = url;
  
    // Check if it's from localhost (http only)
    const isLocalhost = protocol === 'http:' && hostname.startsWith('localhost');
  
    // Check if it's from a Braze domain (must be https)
    const isBrazeDomain =
      protocol === 'https:' &&
      (hostname.endsWith('.braze.com') || hostname.endsWith('.braze.eu'));
  
    if (!isLocalhost && !isBrazeDomain) return;
  
    // Process the message
    const { type, payload } = event.data || {};
    if (type === 'BRAZE_DEVELOPER_DATA' && payload) {
      window.localStorage.setItem('braze_developer_data', JSON.stringify(payload));
    }
  } catch (e) {
    // event.origin might not be a valid URL
    console.error('Invalid origin:', event.origin);
  }
});
```

zzhaobraze (approved):
I have no issues with this if the stakeowners are fine with this but i'll let @josh-mccrowell-braze make final approval

josh-mccrowell-braze (approved):

PR #9443, opened 2025-04-10
expand web instructions

davidbielik:
would prefer these to be expanded to make CTRL+F searches work on this page


internetisaiah (approved):
LGTM, i didn't realize this blocked cmd+f, my bad!

PR #9572, opened 2025-04-29
add language cleanup script

zzhaobraze:
Cleans up any language file that's not an english docs file due to being deleted or moved.

# Usage
```
This script finds and deletes all files that exist in one or more '_lang' subdirectories
that don't exist in the primary '_docs' directory. It can be run against a single language or all languages.

USAGE:
  ./scripts/clean_orphaned_translations.py [LANGUAGE]

OPTIONS:
  [LANGUAGE]         Process the given language. If none, process all. Available languages:
                       ja, de, es, fr, ko, pt-br
  -h, --help         Show this help message
```

# Output
```
===== SUMMARY OF CLEANED FILES =====

ES - Total: 2 files cleaned
Files cleaned by section:
  - _developer_guide: 2

FR - Total: 0 files cleaned
No files cleaned
```

### DIFF
--- a/scripts/clean_orphaned_translations.py
+++ b/scripts/clean_orphaned_translations.py
@@ -1,0 +1,120 @@
+#!/usr/bin/env python3
+# Usage:  python3 scripts/clean_orphaned_translations.py
+# or:     python3 scripts/clean_orphaned_translations.py [language]
+# Example: python3 scripts/clean_orphaned_translations.py es
+import os
+import shutil
+import sys
+from pathlib import Path
+
+# Base directories
+docs_dir = Path("_docs")
+lang_dirs = {
+    "ja": Path("_lang/ja"),
+    "es": Path("_lang/es"),
+    "fr": Path("_lang/fr_fr"),
+    "ko": Path("_lang/ko"),
+    "pt-br": Path("_lang/pt_br")
+}
+
+def get_all_files(directory):
+    """Get all files in a directory recursively."""
+    all_files = []
+
+    for root, dirs, files in os.walk(directory):
+        for file in files:
+            rel_path = os.path.relpath(os.path.join(root, file), directory)
+            all_files.append(rel_path)
+
+    return all_files
+
+def main():
+    # Get all files in the _docs directory
+    docs_files = set(get_all_files(docs_dir))
+
+    # Initialize a dictionary to track orphaned files by section for each language
+    language_summaries = {}
+
+    # Check if a specific language was specified via command line
+    target_language = None
+    
+    if len(sys.argv) > 1:
+        arg = sys.argv[1]
+        if arg in ('--help', '-h'):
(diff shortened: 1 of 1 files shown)


zzhaobraze:
> @zzhaobraze @josh-mccrowell-braze we're missing German

thanks added

scripts/clean_orphaned_translations.py
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env python3
+# Usage:  python3 scripts/clean_orphaned_translations.py
```
internetisaiah, line 2: since you have `#!/usr/bin/env python3`, i believe usage is just:

```sugggestion
# Usage: ./scripts/clean_orphaned_translations.py [LANG]
#  If no argument is given, all languages are processed.
```

if i'm correct here, i'll push a commit to cleanup the `--help` message 👍🏽

scripts/clean_orphaned_translations.py
```diff
@@ -0,0 +1,121 @@
...
+# or:     python3 scripts/clean_orphaned_translations.py [language]
```
internetisaiah, line 3:
```suggestion
```

scripts/clean_orphaned_translations.py
```diff
@@ -0,0 +1,121 @@
...
+        for lang_file in lang_files:
+            # Skip files in _includes directory
+            if lang_file.startswith("_includes/"):
+                continue
+
+            # Check if the file exists in _docs
+            if lang_file not in docs_files:
+                orphaned_files.append(lang_file)
+                file_path = os.path.join(lang_dir, lang_file)
+                print(f"Orphaned file found: {file_path}")
```
internetisaiah, line 82: i tested with `ja` and it works great, but with that 1 lang over 800 files were cleaned up, which printed more messages to my terminal than i could scroll up to read.

the final summary is extremely helpful, but what if the line-by-line items are written to a file in:

```
./scripts/temp/cleaned_lang_files.log
```

or something?

internetisaiah (approved):
LGTM. script works, just a couple small suggestions 👍🏽

scripts/clean_orphaned_translations.py
internetisaiah: i fixed this in latest push

scripts/clean_orphaned_translations.py
internetisaiah: i fixed this in latest push

zzhaobraze: not sure why you want to remove this usage line.

internetisaiah:
FYI, going to test this with Iara before merging to ensure the script is deleting the appropriate files. Please don't merge until then 👍🏽

internetisaiah (approved):
LGTM, tested with iara and all seems g2g 👍🏽

PR #9662, opened 2025-05-14
fix broken redirect when destination is an array

zzhaobraze:
Fix issue with redirect when the destination is an array instead of a string.


josh-mccrowell-braze (approved):
Works for me! I suspect those redirects weren't intended to be arrays, just copy/pasting errors on the part of the creators. Thanks for the fix, Z!

PR #9795, opened 2025-06-04
Adjust feedback page

zzhaobraze:
Change feedback stars from ul to radio and some formatting changes


bre-fitzgerald:
Thanks @zzhaobraze! This is on my to-do list to review later today 🙏

bre-fitzgerald:
@zzhaobraze thanks so much for making these adjustments! The styling changes look good to me overall, there are just two small areas for your review:

**Keyboard focus order**
The rating stars are in a reverse tab order. Tabbing starts at #5 and goes to #1. Shift tabbing goes from #1 to #5. This is the opposite of expected behavior. Can we make it so that tabbing starts at the first star, and proceeds logically from left to right? See  video for problem behavior.

https://github.com/user-attachments/assets/9b36fb11-7f46-42a5-9595-c18f6707f9cc


**CTA styling**
To make our buttons a little less visually overwhelming, can we apply the styling changes outlined here? Currently this page has two primary CTAs (prominent black buttons, which draw the eye). The "Start contributing" is a secondary CTA and should have less visual weight. 

image

bre-fitzgerald:
@zzhaobraze this looks good to me, it just appears there's a minor merge conflict 👀

PR #10021, opened 2025-07-11
fix issue with include path not correctly working for fr_fr and pt_br

zzhaobraze:
Fix issue with fr and pt_br not point to the correct includes


rachel-feinberg (approved):
LGTM!

PR #10024, opened 2025-07-14
File new_features.md committed.

Translationmanagement:


internetisaiah (approved):
@iaraltkorn , fyi had to fix a `{% multi_lang_include %}` in this commit, but now it's all good to go.

maybe i can take a look at our existing regex's too and see if there's any tweaks i can make as well.

PR #10060, opened 2025-07-16
change ja support topic to be english

zzhaobraze:
- Add option for the select text and value to be different.
- Changed ja feedback to use english topic values


internetisaiah (approved):
tested locally in `ja` skew. everything seems to be working properly from my limited testing ✅ 

Screen Shot 2025-07-16 at 12 57 54 PM

PR #10069, opened 2025-07-17
add platform and tool to header tags

zzhaobraze:
Adds platform and tools as meta data from the yaml for search indexing


internetisaiah (approved):
LGTM

PR #10284, opened 2025-08-08
Fixed cookie banner on localhost

davidbielik:


_includes/header.html
```diff
@@ -10,7 +10,11 @@
 
 {% comment %} End OneTrust {% endcomment %}
 <!-- OneTrust Cookies Consent Notice start for braze.com -->
-<script src="https://cdn.cookielaw.org/scripttemplates/otSDKStub.js" data-document-language="true" type="text/javascript" charset="UTF-8" data-domain-script="a4189af1-fcc1-4ec1-9f51-e3bc768c6650"></script>
+{%- assign domain_script = "a4189af1-fcc1-4ec1-9f51-e3bc768c6650" -%}
+{%- if jekyll.environment == "development" -%}
+  {%- assign domain_script = domain_script | append: "-test" -%}
```
zzhaobraze, line 15: where did this "-test" come from? also don't think we set the JEKYLL_ENV, anywhere so this will always be blank. I can add JEKYLL_ENV = production to vercel if you want to that env

davidbielik: it's what i see in their docs code examples, i'm verifying that now though.

is there a better way to target development mode?

zzhaobraze: added JEKYLL_ENV=production to the vercel env for production deploy so you can use this. 
are you setting the JEKYLL_ENV=development locally for this? ie `export JEKYLL_ENV=development && rake` might be better to do `{%- if jekyll.environment != "production" -%}`

davidbielik: image

confirmed `-test` is correct. 

> added JEKYLL_ENV=production to the vercel env for production deploy so you can use this.
are you setting the JEKYLL_ENV=development locally for this? ie export JEKYLL_ENV=development && rake might be better to do {%- if jekyll.environment != "production" -%}

I believe it is `development` by default but i'll change it

zzhaobraze (approved):

davidbielik:
> > @zzhaobraze does it reappear on reload though? that's what this fixes.
> > Also the `-test` is intentional / in their docs for localhost sites.
> 
> yes it still shows on refresh, you can see it in the preview site https://braze-docs-dxzifco76-braze.vercel.app/docs/

This is for localhost (`127.0.0.1:4000`). I don't think we can fix it on deployment hosts unless we comment it out completely.

atom-evens:
> @atom-evens did you test with a different browser? if so, it fine to merge

@zzhaobraze, I texted in Chrome and Safari, and all good on my end. I'll go ahead and approve and merge, but let me know if there's anything else!

PR #10172, opened 2025-07-30
[SDK-5974] [BD-5080] Banner Properties

jacksonemiller:
### Why are you making this change? (required)

To document the new public API methods added to Banner objects to access the properties/KVPs you will soon be able to associate with them.

This PR also adds some presently missing code snippets for Swift to a feature flag tutorial involving properties/KVPs.

I was thinking we could use this PR for both the Android and Swift changes. Then maybe make a separate branch for wrappers? Or we could just use this one for all of them.

Anyway, I'm ready to put the Swift changes into review, so I am going to request review from those folks.

### Related PRs, issues, or features (optional)

- [BD-5080](https://jira.braze.com/browse/BD-5080)

### Feature release date (optional)

- N/A

### Contributor checklist

- [ ] I confirm that my PR meets the following:
    - I signed the [Contribution License Agreement (CLA)](https://www.braze.com/docs/cla).
    - My style and voice follows the [Braze Style Guide](https://docs.google.com/document/u/2/d/e/2PACX-1vTluyDFO3ZEV7V6VvhXE4As_hSFwmnFFdU9g6_TrAYTgH1QmbRoEDDdn5GzKAB9vdBbIdyiFdoaJcNk/pub).
    - My content contains correct spelling and grammar.
    - All links are working correctly.
    - If I renamed or moved a file or directory, I set up [URL redirects](https://www.braze.com/docs/contributing/content_management/redirecting_urls) for each file.
    - If I updated or replaced an image, I did not remove the original image file from the repository. (For more information, see [Updating an image](https://www.braze.com/docs/contributing/content_management/images/#updating-an-image)).
    - If my PR is related to a paid SKU, third party, SMS, AI, or privacy, I have received written approval from Braze Legal.

### Submitting for review

If your PR meets the above requirements, select **Ready for review**, then add a reviewer:
- **Non-Braze contributors:** Tag `@braze-inc/docs-team` in a comment below.
- **Braze employees:** Have any relevant subject matter experts (like engineers or product managers) review your work first, then add the [tech writer assigned to your specific vertical](https://confluence.braze.com/pages/viewpage.action?pageId=75039896). If you're not sure which tech writer to add, you can add `braze-inc/docs-team` instead. 

Thanks for contributing! We look forward to reading your work.

PR #10390, opened 2025-08-21
Update creating_campaigns.md

mark-at-braze:
added custom property info

will follow up offline re: when to merge this


niccrowley (approved):

internetisaiah:
moving this content over to #10172 10172

PR #10583, opened 2025-09-18
File talonone.md committed.

Translationmanagement:


_lang/fr_fr/_partners/data_and_analytics/business_intelligence/clarisights.md
```diff
@@ -0,0 +1,70 @@
+---
```
internetisaiah, line 1: example

_lang/fr_fr/_partners/data_and_analytics/business_intelligence/thoughtspot.md
```diff
@@ -0,0 +1,42 @@
+---
```
internetisaiah, line 1: example

internetisaiah (changes requested):
looks like this PR completely annihilated all content in the `_lang/fr_fr` dir, so closing it out. i double-checked the regex and everything _seems_ fine in phrase, so i'm not sure how to fix it.

but it will likely need to be re-triggered.

